PR #44, opened 2014-03-29
add close unmodified context menu

gschrader:
My common workflow is poking around in a bunch of other files to figure something out and then want to get back and focus only on what I've changed.
close-unmodified


simon04:
May I comment on this old issue w.r.t. "Close Unmodified".

IntelliJ IDEA also calls it "Close Unmodified". However other than Atom, IntelliJ automatically saves files after any modification. So there's no ambiguity in the label.

Anyhow this is my favourite closing mode to quickly get rid of all (no longer) relevant tabs.

Issue #109, opened 2015-01-24
Suggestion: Hide tab bar when only 1 tab is open


modest:
Would be a nice config option, at least.  This follows the behavior of many other tabbed Mac apps such as Safari.

![screen shot 2015-01-24 at 3 30 04 pm](https://cloud.githubusercontent.com/assets/990299/5889467/feea1cc4-a3dd-11e4-93b4-a67a9d375b62.png)
![screen shot 2015-01-24 at 3 30 24 pm](https://cloud.githubusercontent.com/assets/990299/5889468/0200318c-a3de-11e4-8535-57abf33e2ac0.png)

simurai:
@modest Are you thinking of changing this behaviour for all themes? Or just the one in your screenshot, I believe it's Atom Light UI.

The [Unity UI](https://atom.io/themes/unity-ui) theme does it like that:

![screen shot 2015-01-25 at 10 48 01 am](https://cloud.githubusercontent.com/assets/378023/5889853/a95f842c-a47f-11e4-8ad3-5eacbe5426b2.png)

And maybe more. It's probably better to leave it up to the theme to decide.

benogle:
Closing as this is a theme related issue. Unity ui does what you would like.

modest:
Sorry, screenshot was misleading.  Replaced with the right before & after picture.  Unity is unrelated.

benogle:
This is unlikely to be added to core. Maybe one could create a theme that hides the tab bar when there is only one tab.

simurai:
hey, "there is a package for that™": https://atom.io/packages/one-tab

ps. had reload Atom after installing, but after it works fine

modest:
Since this is fairly standard default behavior across tabbed apps (see: TextMate, Finder, Safari, Transmit), it might be reasonable to include this in the core tabs package, at least as a setting :)  Gave it my best shot.

PR #111, opened 2015-01-27
Add support for hiding bar when ≤1 tab (#109)

modest:
Introduces user setting "Hide Tab Bar When Only One Tab Is Open" (#109)

screen shot 2015-01-26 at 10 53 37 pm

screen shot 2015-01-26 at 10 54 21 pm

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -51,7 +51,7 @@ class TabBarView extends View
 
     @subscriptions.add atom.config.observe 'tabs.tabScrolling', => @updateTabScrolling()
     @subscriptions.add atom.config.observe 'tabs.tabScrollingThreshold', => @updateTabScrollingThreshold()
-    @subscriptions.add atom.config.observe 'tabs.hideTabBarWhenOnlyOneTabIsOpen', => @updateSingleTabBehavior()
+    @subscriptions.add atom.config.observe 'tabs.hideTabBarWhenOnlyOneTabIsOpen', => @updateTabBarVisibility()
 
     @updateActiveTab()
 
@@ -113,7 +113,7 @@ class TabBarView extends View
     @updateTabBarVisibility()
 
   updateTabBarVisibility: ->
-    if @hideTabBarWhenOnlyOneTabIsOpen and not @shouldAllowDrag()
+    if atom.config.get('tabs.hideTabBarWhenOnlyOneTabIsOpen') and not @shouldAllowDrag()
       @element.classList.add('hidden')
     else
       @element.classList.remove('hidden')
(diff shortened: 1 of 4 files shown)


50Wliu:
If this is the default for most Mac programs, then I guess the default could be changed to `process.platform is 'darwin'`, right?

lib/tab-bar-view.coffee
```diff
@@ -104,12 +105,20 @@ class TabBarView extends View
     else
       @element.appendChild(tab)
     tab.updateTitle()
+    @updateTabBarVisibility()
 
   removeTabForItem: (item) ->
     @tabForItem(item)?.destroy()
     tab.updateTitle() for tab in @getTabs()
+    @updateTabBarVisibility()
     return
```
benogle: This return can be removed because we no longer have a look at the end of the function

benogle:
Looks legit, i'd rather leave the default to false rather than true on macs.

modest:
I'm fine with defaulting to false on OS X.  Even though some major Mac apps follow this behavior (e.g. Finder, Safari, TextMate, Transit), there are others that choose to always show tabs (Chrome, Firefox).

There are also minor limitations to functionality when the tab bar is hidden, mainly around drag & drop:
* A tab in a 1-tab window cannot be moved to another window.
  * Same limitation exists in the implementations mentioned above.
* A tab from another window cannot be moved into a 1-tab window.
  * This limitation exists because the only drop target for tabs is the tab bar (in Atom).  Unlike other apps, Atom does not support dropping tabs onto the main window area.

So default=false is reasonable.

lib/tab-bar-view.coffee
```diff
@@ -315,6 +324,10 @@ class TabBarView extends View
     else
       @off 'wheel'
 
+  updateSingleTabBehavior: ->
+    @hideTabBarWhenOnlyOneTabIsOpen = atom.config.get('tabs.hideTabBarWhenOnlyOneTabIsOpen')
```
kevinsawicki: Might be worth it to just check `atom.config.get('tabs.hideTabBarWhenOnlyOneTabIsOpen')` in the `updateTabBarVisibility` if check, this instance variable seems only used there so it could then this line could be removed.

modest: Fixed

lib/tab-bar-view.coffee
modest: Fixed

kevinsawicki:
Cool, thanks for making the updates, this will be included in the next Atom release, 0.177

simurai:
Instead of

screen shot 2015-01-28 at 11 58 53 am

should it use a shorter title and add a longer description? Something like:

screen shot 2015-01-28 at 11 59 08 am

modest:
That's a reasonable option. The current name is clear but admittedly verbose.

What about inverting it and describing the positive state?  This feels vague familiar:

"[√]  Always Show Tab Bar
Shows the Tab Bar when only 1 tab is open"

benogle:
I like that verbiage @modest

PR #118, opened 2015-02-06
Add "data-type" attribute to tabs

simurai:
This PR adds a `has-path` class for tabs with a path. This will allow themes to style `<atom-text-editor>` tabs differently than tabs like the Settings or Deprecation Cop etc.

A use case would be to make active tabs have the same (seamless) color as a Syntax theme's background, but tabs like for the Settings, keep the UI colors. See https://github.com/atom/one-dark-ui/pull/11#issuecomment-73009941

screen shot 2015-02-05 at 4 51 31 pm

ps. Not sure if there is a better way to tell if it's a `<atom-text-editor>` tab, but using `if itemPath = @item.getPath?()` seems to work.

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -5,8 +5,6 @@ module.exports =
 class TabView extends HTMLElement
   initialize: (@item) ->
     @classList.add('tab', 'sortable')
-    if itemPath = @item.getPath?()
-      @classList.add('has-path')
 
     @itemTitle = document.createElement('div')
     @itemTitle.classList.add('title')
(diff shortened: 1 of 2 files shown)


kevinsawicki:
I'm not sure the present of a path is enough since if you create a new empty buffer, it has no path.

I pushed a commit to this PR that instead adds the item's constructor name as a data attribute.

So you should be able to select text editors using this data attribute.

screen shot 2015-02-09 at 9 55 27 am

What do you think?

simurai:
:+1:  Yessss, having a `data-type` is much better. Then you *could* target other things like the settings too.

PR #119, opened 2015-02-12
Feature/shift click close tab

peterlozano:
I find very confortable that my current editor (phpstorm) allows to close tabs using shift + click.

I've implemented this because:
* I don't have a middle button
* I find it faster than clicking on the close button.
* As my first exercise to hack atom :)

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -75,8 +75,11 @@ class TabBarView extends View
       @pane.destroyItem(tab.item)
       false
 
-    @on 'click', '.tab', ({target, which}) =>
+    @on 'click', '.tab', ({target, which, shiftKey}) =>
       tab = $(target).closest('.tab')[0]
+      if which is 1 and shiftKey
+        @pane.destroyItem(tab.item)
+        return false
       if which is 1 and not target.classList.contains('close-icon')
         @pane.activateItem(tab.item)
         @pane.activate()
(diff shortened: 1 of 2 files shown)


50Wliu:
:-1: I disagree with this.  `Shift+Click` in Chrome allows you to select multiple tabs at the same time (and while that isn't implemented in Atom yet, I hope it is soon) and it's going to be pretty confusing if someone tries to do that and bam, tab closed.  As a workaround, `Ctrl+W` is always available.

peterlozano:
It wasn't confusing for me when I found that feature in phpstorm, in fact I find it very convenient and time saving and I miss it in Atom, it is very useful for closing tabs your are no longer using. (ctrl+w closes the current tab, shift+click can close any tab)

I can't hardly think of a use case where selecting multiple tabs was not for closing them. Actually I think the shift+click multiple selection is a feature I have never used (or missed) in Chrome or any other software.

Some questions that cross my mind:
* Could this be made a setting?
* Could this be implemented from an external package? I still don't know the api well. Can the click handler be overriden somehow?
* Can the metrics/analytics shed some light on what the users are expecting?
* Are there other feature request for the shift+click multiple tab selection functionality from other people?

lee-dohm:
I'm pretty sure it could be implemented from an external package. The tabs package doesn't really do much but provide visualization for the pane/editor structure that already exists in the base API.

benogle:
Closing as this seems like a great idea for a package

PR #120, opened 2015-02-18
Push top-level require() calls down into the functions that use them.

bolinfest:


izuzak:
Look what I found! Thanks, @bolinfest :star2:

Issue #123, opened 2015-03-16
Switch tabs onMouseDown


KeitIG:
Hello,

I was surprised to not find any issue on this. What about switch tabs onMouseDown() instead of Click() ? It has all benefits: you can see the tab sooner, the UI seems more reactive...

ref: https://github.com/atom/atom/issues/5174

As as-cii mentioned, it is not really "pure performance improvement". Just a (working) trick.

I guess I could do a PR myself, but what do you think about that ?

swinbank:
I do agree that perceived responsiveness is important. I tinkered with this about a year ago, but I vaguely remember it being slightly more trivial than I had bargained for, since it was affecting the ability to drag tabs around.

KeitIG:
@swinbank take a look at my PR. I've fixed it ;)

swinbank:
Great!

50Wliu:
This has been merged so closing.

PR #124, opened 2015-03-16
Switch tabs onmousedown

KeitIG:
following #123 

I tested it, It works quite well and feels more reactive. Merge if you feel the need.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -60,7 +60,7 @@ class TabBarView extends View
       if which is 1 and not target.classList.contains('close-icon')
         @pane.activateItem(tab.item)
         @pane.activate()
-        false
+        true
       else if which is 3 or (which is 1 and ctrlKey is true)
         @find('.right-clicked').removeClass('right-clicked')
         tab.classList.add('right-clicked')
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -75,7 +75,7 @@ class TabBarView extends View
       @pane.destroyItem(tab.item)
       false
 
-    @on 'click', '.tab', ({target, which}) =>
+    @on 'mousedown', '.tab', ({target, which}) =>
       tab = $(target).closest('.tab')[0]
       if which is 1 and not target.classList.contains('close-icon')
         @pane.activateItem(tab.item)
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -57,7 +57,11 @@ class TabBarView extends View
 
     @on 'mousedown', '.tab', ({target, which, ctrlKey}) =>
       tab = $(target).closest('.tab')[0]
-      if which is 3 or (which is 1 and ctrlKey is true)
+      if which is 1 and not target.classList.contains('close-icon')
+        @pane.activateItem(tab.item)
+        @pane.activate()
+        false
+      else if which is 3 or (which is 1 and ctrlKey is true)
         @find('.right-clicked').removeClass('right-clicked')
         tab.classList.add('right-clicked')
         false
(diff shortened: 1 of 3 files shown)


thomasjo:
@KeitIG You've got a broken spec which needs to be fixed:
https://travis-ci.org/atom/tabs/builds/54543728#L111

KeitIG:
@thomasjo 

I got the error, (conflict between two mousedown events). Is it possible to test Travis builds without committing to avoit commits pollution ?

KeitIG:
now done. 630883e was more a code cleanup than a spec fix.

as-cii:
@KeitIG thank you for taking the time to contribute! :bow: 

I quickly tested it and, apparently, dragging tabs is not working anymore. My first concern is that specs are not failing, probably because we fake the drag event when testing this behavior.

Moreover, UX-wise, I do have one more concern: what would happen if I simply want to drag a tab which is not currently open? This way we would always open the tab, even if the user didn't want to. Personally, I would find this a bit confusing but I am not sure it's wrong either.

@kevinsawicki: what do you think?

KeitIG:
@as-cii Yep, got the bug too. I may have an idea about why.

On the UX part, looking at Sublime, imagine you'are at `tabs[0]`, if you want to place `tabs[1]` before `tabs[0]`, `tabs[1]` is opened when it gets the mousedown event.

I'd be curious what other people are thinking about this.

50Wliu:
@as-cii This is how Chrome does it, but it does feel a bit weird when you didn't actually want the tab to open.

simurai:
I like it. It feels a bit more "effortless".

> @as-cii what would happen if I simply want to drag a tab which is not currently open? This way we would always open the tab, even if the user didn't want to.

Currently if you drag a not opened tab, it still opens that tab on `mouseup`. So it's currently not possible to drag without opening.

Not sure which is the better behaviour. Sometimes you want to drag + edit a tab and therefore having it open without an extra click, is nice too.

as-cii:
> Currently if you drag a not opened tab, it still opens that tab on mouseup. So it's currently not possible to drag without opening.

@simurai: nice observation! For some reason I assumed that the `mouseup` event was not propagated whenever a drag occurred. Since we're basically preserving the original behavior I think we can merge this after refining it a bit.

@KeitIG: I [opened another pull-request](https://github.com/KeitIG/tabs/pull/2) on your fork to shield us from the dragging regression in the future.

As I described there, writing an integration test (e.g. through a web driver) was not an option, and simulating the dragging events through a series of `mousedown`, `mousemove` and `mouseup` events was not really straightforward as we had to deal with positions. Therefore, I opted to go with a unit test which ensures that `event.preventDefault` gets called only when appropriate.

/cc: @kevinsawicki

kevinsawicki:
So this is ready to go?

KeitIG:
I guess so, as-cii have done a nice job. (Thanks btw, I would not have figured it myself)

as-cii:
@kevinsawicki: yep! I have just merged it :+1: 

@KeitIG: thank you for this great contribution! :sparkles:

izuzak:
@as-cii @KeitIG I _think_ this PR broke some focus-related things when switching tabs -- see https://github.com/atom/atom/issues/6662 for details. Specifically, I think https://github.com/atom/tabs/commit/6cb57b8b8fda3258cde6b2e26a673e5c2368e49d might be responsible. Could you take a look and double-check? :doughnut:

KeitIG:
Indeed, It may be due to that, looking for a way to rewrite this.

nathansobo:
@KeitIG You might try calling `activate` on the associated pane at the model level once a new pane is selected. That should instruct the pane view to focus its contents again and might work. I don't have deep insight into what's going on but that could be a helpful tool.

nathansobo:
Nevermind, I see activate is already being called and somehow focus is still getting stolen.

nathansobo:
I've opened a PR which you can see above which fixes the issue for me. @izuzak want to give it a spin and see if it breaks anything else. I couldn't find anything but I'm a sucky QA tester.

Issue #39, opened 2014-03-21
Pop out tabs to create new editor window


chamons:
In many "tab" based products, you can grab a tab and "pop it out" to create a new window with that content. That would be awesome for atom when you want to windows side by side.

halleycarleton:
+1 Would love to see this implemented

afrjorge:
+1

jaumann:
+1

daveqr:
+1

bart452:
+1

adcreare:
+1

luiscarrasc0:
+1

Platzer:
+1

Chesire:
+1

markuz-brasil:
Are you fucking kidding me. 
Like this is such a early issue and no official response.

like, how hard would that be to implement

benogle:
> like, how hard would that be to implement

Thanks for the comment. PRs would be accepted for such a feature.

joeyguacamole:
+1

aveiros:
+1

jy19:
+1

bcherny:
+1

davidgross:
I have been fiddling with this for the last couple days, (I really want this feature), I am not a node/coffee/javascript programmer(I am a Rails dev, I have some limited javascript experience....).  I have a couple questions for someone who knows something about how the events work for tabs, I figured out how to make it make a new window when you drag a tab out of the window(using onDragLeave).

The trouble is, it creates it when you drag it within the window as well, and if you drag it over several tabs(eg: re-ording tabs) it will open a bunch of new windows up.

It seems like I need a way to determine if you dropped it outside the window, but there does not seem to be an event for that(or I don't know what it is called)

onDragEnd seemed promising, but it doesn't allow you to drop the tab in a new window, as it creates the new window after you drop the tab....

 I am using the "application:new-window" command to get this, which means you also end up with a "untitled" blank-tab, and that seems wrong...

Thoughts?

PS: my tweaks are in https://github.com/davidgross/tabs if anybody cares....(also, my stuff really just breaks tabs at the moment....)

manuelsalvatori:
+1

davidgross:
Update on my tweaks: I have a functioning "Open Tab in new Window" option in the tab right click menu.

This will, as the name suggests, open the tab in a new window, _however_ it won't copy your un-saved changes, but it does prompt you to save before opening the file in a new window.

[...]

PR #134, opened 2015-04-27
Add "open in new window" option to tabs context menu

davidgross:
This creates a workaround for https://github.com/atom/tabs/issues/39

I have added a context menu option for opening a tab in a new window.

I placed it in the "reveal in treeview, reopen closed tab" section, below the split section. 

It just spawns a new atom session, and then opens the file in question, and closes it in the current(old) window.

It will prompt the user to save the file before closing the tab in the old window, if there are unsaved changes.

tab-bar-view coffee - -home-grosda-work-atom-tabs - atom_003


menus/tabs.cson
```diff
@@ -12,6 +12,10 @@
     {label: 'Split Down', command: 'tabs:split-down'}
     {label: 'Split Left', command: 'tabs:split-left'}
     {label: 'Split Right', command: 'tabs:split-right'}
+    
+    {type: 'separator'}
+    
+    {label: 'Open this tab in new window', command: 'tabs:open-tab-in-new-window'}
```
mnquintana: Would you mind title-casing this to match the other commands?

davidgross: Yep, done.

davidgross:
The other thing not in this, is a jasmine test to confirm that the button actually works.  I have not written tests with that before, it seems that the open window button is not tested in the tree view, which is what I based my "Open Tab In New Window" button on.  And a grep in the atom repo didn't result in anything that helped me either(it may be there, I just didn't find it).

davidgross:
Oh, and the name of the button, it feels poorly worded to me, but it was the best I could come up with. If there is a more succinct way of working it, I am happy to change it.

lee-dohm:
@davidgross I think you can just call it "Open In New Window", since they have already clicked the tab to get the context menu, yes? That would be the same title as is used when right-clicking a file in the Tree View.

davidgross:
@lee-dohm Yeah, that makes sense, doing that now.

lib/tab-bar-view.coffee
```diff
@@ -24,6 +24,7 @@ class TabBarView extends View
       'tabs:split-down': => @splitTab('splitDown')
       'tabs:split-left': => @splitTab('splitLeft')
       'tabs:split-right': => @splitTab('splitRight')
+      'tabs:open-tab-in-new-window': => @openTabInNewWindow()
```
mnquintana: Would you mind renaming this to `tabs:open-in-new-window` to match the new context menu entry name?

lib/tab-bar-view.coffee
```diff
@@ -135,6 +136,18 @@ class TabBarView extends View
   closeTab: (tab) ->
     tab ?= @children('.right-clicked')[0]
     @pane.destroyItem(tab.item)
+    
+  openTabInNewWindow: (tab) ->
```
mnquintana: Would you mind renaming this to `openInNewWindow` to match the new context menu entry name?

mnquintana:
Other than the one comment, would you mind adding some specs for this?

davidgross:
@mnquintana Sorry...apparently I am not getting email about this, I just happened to check this page, and see the comments.  

Do you mean "specs" as in the tests? or specs as in more detail on what this does? ( I assume the tests?)

I changed the tabs command name.

mnquintana:
@davidgross Sorry, yup, specs meaning tests. Thanks!

davidgross:
@mnquintana tests added, hopefully they work like I think they do.

lib/tab-bar-view.coffee
```diff
@@ -135,6 +136,18 @@ class TabBarView extends View
...
+    tab ?= @children('.right-clicked')[0]
+    item = tab.item
+    @closeTab(tab)
+    if typeof item.getURI is 'function'
+      itemURI = item.getURI() ? ''
+    else if typeof item.getPath is 'function'
+      itemURI = item.getPath() ? ''
+    else if typeof item.getUri is 'function'
+      itemURI = item.getUri() ? ''
+    atom.open({pathsToOpen: [itemURI], newWindow: true})
```
jerone: A tab could have non of above functions (`getURI`, `getPath`, `getUri`). Wouldn't it better to check if `itemURI` has a value before closing the existing tab and opening a new window with maybe empty paths...

cc @mnquintana

davidgross: @jerone did you mean something like this?
```javascript
openInNewWindow: (tab) ->
    tab ?= @children('.right-clicked')[0]
    item = tab.item
    return unless item?
    if typeof item.getURI is 'function'
      itemURI = item.getURI() ? ''
    else if typeof item.getPath is 'function'
      itemURI = item.getPath() ? ''
    else if typeof item.getUri is 'function'
      itemURI = item.getUri() ? ''
    return unless itemURI?
    atom.open({pathsToOpen: [itemURI], newWindow: true})
    @closeTab(tab)
```

jerone: [**@davidgross**](https://github.com/davidgross) commented on [12 mei 2015 20:10 CEST](https://github.com/atom/tabs/pull/134/files#r30162873 "2015-05-12T18:10:06Z - Replied by Github Comment Enhancer"):
> [@jerone](https://github.com/jerone) did you mean something like this?

Yeah.

jerone: The `? ''` might be redundant then.

davidgross: ok, I stole that from onDragStart.....

davidgross:
The test is failing, but the function works, correctly. 
For some reason, when I call: 
```javascript
atom.commands.dispatch(tabBar.element, 'tabs:open-in-new-window')
```
The element is not valid.

Any thoughts?

jerone:
[**@davidgross**](https://github.com/davidgross) commented on [13 mei 2015 01:31 CEST](https://github.com/atom/tabs/pull/134#issuecomment-101459705 "2015-05-12T23:31:12Z - Replied by Github Comment Enhancer"):
> The test is failing, but the function works, correctly.   
> 
> For some reason, when I call: 
> 
> <div class="highlight highlight-javascript"><pre>atom.commands.dispatch(tabBar.element, <span class="pl-s"><span class="pl-pds">'</span>tabs:open-in-new-window<span class="pl-pds">'</span></span>)</pre></div>
> 
> The itemURI is not valid.
> 
> I see no reason that would be the case, and in actual practice, itemURI is set in every case I can think of to try(manually).
> 
> Any thoughts?

Because the [earlier mentioned situation](https://github.com/atom/tabs/pull/134#discussion_r30155513) is active. The [`TestView`](https://github.com/davidgross/tabs/blob/master/spec/tabs-spec.coffee#L49-L70) doesn't have any of the used methods (`getURI`, `getPath`, `getUri`), and thus `itemURI` isn't returning anything. Resulting in a failed test, because no tab is closed.

davidgross:
Oh, right, ok that makes sense.  I have added a getURI to TestView, which then allows the test to _acually_ test that my "open in new window" function is working properly.

mnquintana:
Alright, just tested this locally - a few things:
- Looks like `Open in New Window` only works on file buffers – for the following, it just opens an untitled buffer in a new window:
  - Settings
  - Timecop
  - Styleguide
  - Deprecation Cop
  - Markdown Preview
  - Project Find Results (e.g. from find-and-replace)
- If you try the command on an untitled, unsaved buffer, it doesn't do anything. (Expected: It opens a new window with your unsaved buffer)
- The new window opened from this command doesn't honor Atom's run modes (e.g. dev mode and safe mode). For example, if you run the command from a dev mode window, the window it opens is not in dev mode.

davidgross:
Any pointers on the unsaved buffer thing? (I had worked on that a bit(before this pull request), and was unable to find a solution)

And atom's run modes, how is that determined?

mnquintana:
> Any pointers on the unsaved buffer thing? (I had worked on that a bit(before this pull request), and was unable to find a solution)

You could do something like:

- If it's an unsaved buffer, [store its contents in a variable](https://atom.io/docs/api/v0.199.0/TextEditor#instance-getText), open a new window with an unsaved buffer, and then [set the text of that new unsaved buffer](https://atom.io/docs/api/v0.199.0/TextEditor#instance-setText) to the contents of the previous unsaved buffer.

> And atom's run modes, how is that determined?

`atom.open` has options for dev mode and safe mode: https://atom.io/docs/api/v0.199.0/Atom#instance-open

Hope that helps!

davidgross:
Ok, those make sense, however, as it sits now, I don't know how to get the process id of the new atom instance.

Or am I not understanding how this works?

mnquintana:
@davidgross You shouldn't need the process id – what were you thinking of using that for?

davidgross:
I was thinking I needed the process id so I could tell the new atom process what to put in it's editor window.

Sheridan-Tech:
+1

davidgross:
@Sheridan-Tech sorry...who/what was the +1 for?

davidgross:
I am still working on this, but getting the newly spawned window to behave is proving tricky.

I can't set the new window's tab's editor's content without knowing something about it.

I can _get_ the content from the current tab, that is easy(item.getText()).

But I have not found a way to _set_ the new window's tab's content.

jerone:
Is it really necessary to move unsaved files/buffers to the new window too?

You are already opening all files that have a path, right?

Sheridan-Tech:
I was attempting to convey an up-vote to the feature. I'd like to see this
feature functioning as soon as possible. I thought it was working as is but
the code doest execute correctly on my systems.

davidgross:
@jerone Correct, if the file has been saved, it works, if it has not been saved _recently_ but has a valid path, it will prompt the user to save the file, then it will close the tab, and open the file in the new window.

The reason I was working on the unsaved buffer part was because things like markdown preview don't behave, at all...when you click the "open in new window" button. (also @mnquintana asked for it.)

davidgross:
@Sheridan-Tech What breaks on your systems? What are your systems? How are you launching Atom?  How are you trying to use my code?

Sheridan-Tech:
It removes all tabs in atom. I am using settings sync so they have all the
same settings and customizations. I have tried installing it alongside
tabs, or standalone without tabs, and tried merging the two.

mnquintana:
@Sheridan-Tech It seems unlikely to me that that is caused by this PR – for you to use a locally cloned branch of a package (ie. this PR) you'd have to manually clone it on each machine and manually `apm link -d` it on each machine. As far as I can tell no settings sync package does that for you.

davidgross:
@Sheridan-Tech Woah....Yeah....I don't get that behavior...

What version of atom is this?
What OS are you running on?

Sheridan-Tech:
I had manually cloned it on each machine. I was just presenting the
settings sync as a possible source of problems. The computers have the same
settings so if tabs conflicted with a setting on one of my machines it
would have the same problem on all.

davidgross:
I've been really busy the last month, so I have not touched this in a while.  I merged in tabs/master, cleaned up some lint stuff, and fixed a couple things I broke when I merged.

This thing still works for saved files, but not for un-saved files, or non-existent files(eg: Markdown preview).

As far as I can tell, something in atom itself will have to change before moving unsaved content from one atom instance to another works... The reason drag and drop works, is due to the drag object having the content.  I have yet to figure out how to make that happen for atom.open, or any other method of opening in a new window.  Also, if you have two atom windows open, and try to drag a markdown preview window over, it _will_ fail. 

I guess, my thought on this is, if you want what I have, take it.  If I figure out how to make this behave, I'll make another PR.  Again, it appears to me that to make the open non-existent file in a new window thing work, something in atom itself is going to need to be added/changed.

Thoughts?  @mnquintana @lee-dohm @jerone

colllin:
I don't mean to hijack your PR, but what do you guys think about setting up shared context menu items for file-related actions? To be shared between atom/tabs and atom/tree-view?  Currently the treeview has:

- Split Up
- Split Down
- Split Left
- Split Right
- ...
- Rename
- Duplicate
- Delete
- Copy
- Cut
- Paste
- ...
- Copy Full Path
- Copy Project Path
- Open in New Window
- Show in Finder

And the tabs have:

- Split Up
- Split Down
- Split Left
- Split Right
- Reveal in Tree View

Obviously it doesn't make sense for the tree-view to have a "Reveal in Tree View" option, but I think the rest of the file-related actions make sense in both places (and some of them already exist in both places).

mnquintana:
@colllin Moving that discussion to a new issue really would be best – let's keep the discussion here focused on this PR.

railsjedi:
+1, this feature would be really really useful

babineaum:
:+1:

b-long:
:+1:

breiter:
:+1:

mnquintana:
Please stop +1ing this PR – it hasn't been merged because it's not ready to ship yet, not because we forgot, and +1ing it won't make it ship any faster.

davidgross:
Just close the PR. You guys clearly don't want the feature. It is as ready
as I can make it. I stated in the beginning, this is a work-around.

Totemika:
This'd be too nice :+1:

adam-knights:
@mnquintana is this nearly there? ps thanks for the work on this @davidgross

Jiri-Kremser:
:ship: ship it

webdevbrian:
Coming up on a year gentlemen. I'm here for the 'open tab in new window' functionality specifically. A bit shocked it took this long to at least merge that simple add in, as so many people rely on that feature daily.

:eyes: @mnquintana

lee-dohm:
Thanks @davidgross for all your work on this. If someone wants to volunteer to take the work here and get it across the finish line, here's what I think needs to be done:

* (Required) Don't show "Open in New Window" menu item on tab items that aren't a `TextEditor`
* (Optional) Show "Open in New Window" menu item on tab items that are tested to work, perhaps a whitelist
* (Recommended) See if there is some way to send across the list of `atom.project.getPaths` to the new window

davidgross:
@lee-dohm 

Can we dynamically change the context menu per tab? I see no way to do that, as the context menu is statically generated on atom start.

I do have it alerting if open in new window is clicked for a non TextEditor Tab(alerts, then returns, does nothing)
Right now I am only aware of this working on TextEditor tab items, but it would be simple to update the white list.
I have atom.project.getPaths passing though.

mnquintana:
@davidgross The easiest way to dynamically change the context menu would be to add a class (or some other property) to the relevant DOM element, and scope the context menu items to that class. For example: https://github.com/atom/tabs/blob/master/menus/tabs.cson#L25-L27

davidgross:
Well, look at that....I should have seen that....oh well...

The 'open in new window' option now only shows when item is a TextEditor or TestView(for tests).

lee-dohm:
@davidgross Can you take a look at this?

1. Launch Atom with the tabs repo in it (wherever you have tabs cloned to as the sole project root)
1. Expand the `lib/` directory in the tree view
1. Double-click the `lib/html-helpers.coffee` file in the tree view
1. Right-click the `html-helpers.coffee` tab to get the context menu
1. Select "Open in New Window" in the context menu

**Expected:** The tabs repo to be the only project root

**Actual:** The tabs repo and the lib directory are both project roots

lee-dohm:
Great work on scoping the menu item to text editors! Works perfectly from what I can see :grinning:

davidgross:
So, that actually is a problem in TreeView.  My "Open In New Window" only exists on the tabs themselves. I agree that should be fixed though....

https://github.com/atom/tree-view/blob/4682f49baa54043d7d1a8fe10a8933ef32a80312/lib/tree-view.coffee#L515-L517

has:
```
  openSelectedEntryInNewWindow: ->
    if pathToOpen = @selectedEntry()?.getPath()
      atom.open({pathsToOpen: [pathToOpen], newWindow: true})
```

it would need to have something like this:

```
openSelectedEntryInNewWindow: ->
  if pathToOpen = @selectedEntry()?.getPath()
    pathsToOpen = [atom.project.getPaths(), pathToOpen].reduce ((a, b) -> a.concat(b)), []
    atom.open({pathsToOpen: pathsToOpen, newWindow: true,})
```

lee-dohm:
Hmmmmm ... thanks for that, taking a look.

davidgross:
Oh, hold it, I mis-read what you said, I thought you were using the 'open in new tab' on the TreeView on the left side.

lee-dohm:
Ok cool ... was just typing up some new repro steps :laughing:

davidgross:
Ok, I am reproducing that.

The reason that is happening is I am calling atom.open, and passing two 'pathsToOpen' strings, in an array.
specifically, atom.project.getPaths() and itemURI

see:
https://github.com/davidgross/tabs/blob/master/lib/tab-bar-view.coffee#L161-L162

I am currently seeing if I can make that behavior change....

davidgross:
ok, so it looks like I would need to be able to set the project path separately, and I am not sure I can do that from the atom.open() call.

atom iself or tree-view is trying to be smart, and open the project paths related to the pathsToOpen,

so when I pass both, it sets the project as the atom.project.getPaths(), and then adds another project for the parent directory of the itemURI that I am passing....

Thoughts?

lee-dohm:
Ok, I was afraid of that :grinning: I think a modification to `atom.open` will need to be created. I'll poke around a little bit on this and if everything looks good, I'll merge it tonight.

danielbayley:
> so when I pass both, it sets the project as the atom.project.getPaths(), and then adds another project for the parent directory of the itemURI that I am passing…

I think I have this exact same problem in my [modular-keymaps](https://atom.io/packages/modular-keymaps) package.

lee-dohm:
Thanks for your amazing perseverance @davidgross :grinning:

colepanike:
This is a good workaround, but doesn't really fix #39. Please don't forget about this feature!

NicoHood:
I agree.
I've tested the latest beta and the feature aimed in the PR works great, however it would be more intuitive to drop the tab somewhere else and a now windows will pop up.

davidgross:
I agree with you ( @colepanike and @NicoHood ).

However, because of the limitations in Atom itself(not the tabs plugin) that is not possible(or I could not find a way to make it happen), for now, I keep hoping they will add a mouseDragOutOfWindow type event that will allow drag-and-drop-in-new-window type actions, they have added a lot of stuff to the atom api, I think it is getting closer.

Also, once you have a new window open, you **can** drag tabs between windows(you probably already knew that...but there it is anyway).

NicoHood:
We then should open a new issue for this feature request and then link issue #134 with this feature.

I think you missed my edit while typing your answer. Please reread the 2nd part of my answer :)

davidgross:
@NicoHood indeed :) 

And you are right about that bug, I'll look into that.

davidgross:
@NicoHood 
That is not a new bug, looks like it is un-related to my open-in-new-window method, we should open a separate issue for that, probably(unless there is one, in which case I did not see it in a cursory look a the current issues).

NicoHood:
Go ahead, I think you can give the error description more details than me.

colepanike:
Agree with previous, and also petition that #39 is not solved and should be re-opened.

lee-dohm:
@colepanike #39 isn't closed.

lee-dohm:
@NicoHood #39 isn't closed.

@colepanike @NicoHood #39 isn't closed. The reason it isn't closed is so we don't forget about the feature. Please don't open a new issue that is a duplicate of an existing one, we'll just have to close it.

colepanike:
Ah, so it isn't. Sorry. Should have looked at current state. I confused the referenced thread closure for the closure of the primary thread.

davidgross:
@lee-dohm is there an existing issue for the 'drag-tab-to-other-window-but-not-closing-original-tab' bug?

riclf:
#39 is locked and no longer open for new comments, so it is not possible to comment there.

I am not reading it exactly as I am imaging it so I'd like to take the opportunity to state it here-  On occasion I want to expand my open work on to a second screen. So I'd like to grab an open tab and drag it over to a second screen where it will exist in a new instance of Atom on the second screen. Firefox is a good place to see that working as it should. 

Programmatically this is already nearly being done in that you can drag an open tab in one instance to an open instance of Atom on a second screen. So it seems that the logical extension would be to check if an Atom instance in currently open, if not, open one, and receive the tab.

This UX is in recognizing the multi-screen world devs are working in these days.

davidgross:
@riclf that is what this is a work-around for.  Atom itself(not the tabs plugin) will need some changes to make drag-tab-out-to-new-window work properly.

blah238:
This does not work with Markdown Preview :(

tomchambers2:
I guess its a limitation of this method, but this feature doesn't work if the file is unsaved.

NicoHood:
This does not work if you open a new document (ctrl + n) which is not safed yet. It only works for saved docs.

PR #145, opened 2015-05-22
Added temporary tabs

carlosingles:
Introduces new setting that opens all tabs in a temp state, allowing them to be closed when focus is changed from item to item.

Origin: ddavison/sublime-tabs#43

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -13,10 +13,10 @@ module.exports =
     tabScrollingThreshold:
       type: 'integer'
       default: 120
-    temporaryTabs:
+    useTransientBehavior:
       type: 'boolean'
       default: true
-      description: 'Tabs will close when they lose focus, Tabs can be locked by double clicking the title'
+      description: 'Tabs will only stay open if they are modified or double-clicked'
 
   activate: ->
     @tabBarViews = []
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -13,7 +13,7 @@ module.exports =
     tabScrollingThreshold:
       type: 'integer'
       default: 120
-    useTransientBehavior:
+    usePreviewTabs:
       type: 'boolean'
       default: true
       description: 'Tabs will only stay open if they are modified or double-clicked'
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -13,6 +13,10 @@ module.exports =
     tabScrollingThreshold:
       type: 'integer'
       default: 120
+    temporaryTabs:
+      type: 'boolean'
+      default: true
+      description: 'Tabs will close when they lose focus, Tabs can be locked by double clicking the title'
 
   activate: ->
     @tabBarViews = []
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -48,7 +48,7 @@ class TabBarView extends View
 
     @subscriptions.add @pane.onDidChangeActiveItem (item) =>
       if atom.config.get('tabs.useTransientBehavior') and item instanceof TextEditor
-        if @tab instanceof TabView and @getTabs().length > 1 and @tab.item isnt item and !@tab.keepOpen and @tab.item instanceof TextEditor
+        if @tab instanceof TabView and @getTabs().length > 1 and @tab.item isnt item and not @tab.keepOpen and @tab.item instanceof TextEditor
           @pane.destroyItem @tab.item
         @tab = @tabForItem(item)
       @updateActiveTab()
@@ -92,7 +92,7 @@ class TabBarView extends View
   addTabForItem: (item, index) ->
     tabView = new TabView()
     tabView.initialize(item)
-    if atom.config.get('tabs.useTransientBehavior') and !@tab and item instanceof TextEditor
+    if atom.config.get('tabs.useTransientBehavior') and not @tab and item instanceof TextEditor
       @tab = tabView
     @insertTabAtIndex(tabView, index)
 
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -46,7 +46,11 @@ class TabBarView extends View
     @subscriptions.add @pane.onDidRemoveItem ({item}) =>
       @removeTabForItem(item)
 
-    @subscriptions.add @pane.onDidChangeActiveItem =>
+    @subscriptions.add @pane.onDidChangeActiveItem (item) =>
+      if atom.config.get('tabs.temporaryTabs')
+        if @tab instanceof TabView && @getTabs().length > 1
+          @pane.destroyItem(@tab.item) if !@tab.open
+        @tab = @tabForItem(item)
       @updateActiveTab()
 
     @subscriptions.add atom.config.observe 'tabs.tabScrolling', => @updateTabScrolling()
@@ -88,6 +92,9 @@ class TabBarView extends View
   addTabForItem: (item, index) ->
     tabView = new TabView()
     tabView.initialize(item)
+    if atom.config.get('tabs.temporaryTabs')
+      if !@tab
+        @tab = tabView
     @insertTabAtIndex(tabView, index)
 
   moveItemTabToIndex: (item, index) ->
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -47,8 +47,8 @@ class TabBarView extends View
       @removeTabForItem(item)
 
     @subscriptions.add @pane.onDidChangeActiveItem (item) =>
-      if atom.config.get('tabs.useTransientBehavior') and item instanceof TextEditor
-        if @tab instanceof TabView and @getTabs().length > 1 and @tab.item isnt item and not @tab.keepOpen and @tab.item instanceof TextEditor
+      if atom.config.get('tabs.usePreviewTabs') and item instanceof TextEditor
+        if @tab instanceof TabView and @getTabs().length > 1 and @tab.item isnt item and @tab.isPreviewTab and @tab.item instanceof TextEditor
           @pane.destroyItem @tab.item
         @tab = @tabForItem(item)
       @updateActiveTab()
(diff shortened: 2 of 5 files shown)


carlosingles:
Sorry added an extra commit, forgot that Sublime Text keeps the file open if it has been modified.

benogle:
Multiple commits is totally fine. We never squash.

lib/main.coffee
```diff
@@ -13,6 +13,10 @@ module.exports =
     tabScrollingThreshold:
       type: 'integer'
       default: 120
+    temporaryTabs:
+      type: 'boolean'
+      default: true
+      description: 'Tabs will close when they lose focus, Tabs can be locked by double clicking the title'
```
50Wliu: This is misleading considering that the tabs stay open once they're modified.  Perhaps `Tabs will only stay open if they are modified or double-clicked` or maybe `Tabs with no changes are closed once they lose focus unless they are double-clicked`.

lib/tab-bar-view.coffee
```diff
@@ -46,7 +46,11 @@ class TabBarView extends View
     @subscriptions.add @pane.onDidRemoveItem ({item}) =>
       @removeTabForItem(item)
 
-    @subscriptions.add @pane.onDidChangeActiveItem =>
+    @subscriptions.add @pane.onDidChangeActiveItem (item) =>
+      if atom.config.get('tabs.temporaryTabs')
+        if @tab instanceof TabView && @getTabs().length > 1
```
50Wliu: `and` instead of `&&`

lib/tab-bar-view.coffee
```diff
@@ -46,7 +46,11 @@ class TabBarView extends View
...
+          @pane.destroyItem(@tab.item) if !@tab.open
```
50Wliu: `unless` instead of `if !`

lib/tab-bar-view.coffee
```diff
@@ -88,6 +92,9 @@ class TabBarView extends View
   addTabForItem: (item, index) ->
     tabView = new TabView()
     tabView.initialize(item)
+    if atom.config.get('tabs.temporaryTabs')
+      if !@tab
```
50Wliu: Ditto with the `unless` here (maybe the two if statements can be combined as well?).

lib/tab-view.coffee
```diff
@@ -4,9 +4,17 @@ path = require 'path'
 module.exports =
 class TabView extends HTMLElement
   initialize: (@item) ->
+    @open = false
```
50Wliu: At least to me, `open` sounds like you're opening the tab, not *keeping* it open.  It was especially confusing on line 52 of `tab-bar-view.coffee` where at first I thought that the tab was being closed if it...wasn't open.  Maybe change it to `@keepOpen = false` or `@temporary = true`?

50Wliu:
Also, some specs would be nice for this :smile:.

thomasjo:
> Also, some specs would be nice for this

I'm gonna be a little bit stricter; specs for this is mandatory.

thomasjo:
I don't really like the name `temporaryTabs`. We're in the context of "tabs" already, so maybe it should be something along the lines of `temporaryMode`, `useTransientBehavior`, `transience`, or some such.

Naming stuff is difficult.

carlosingles:
Thanks for the help guys :smiley: 

I've implemented your suggestions and renamed the setting to "useTransientBehaviour" - seems to make the most sense

I also modified the code to prevent TextEditor tabs from being closed when any other panel is opened and focused

I'll work on writing some specs next :+1:

carlosingles:
I'm not extremely familiar with writing specs, so forgive me if they're completely wrong :see_no_evil:

Any tips / help appreciated if I missed anything!

ddavison:
use specs from sublime-tabs.. since this PR is based on that, you should be able to see which tests test which behavior and try accordingly

carlosingles:
Thanks :+1: 

I've updated the specs to use the specs from sublime-tabs and have also updated them to suit this fork.

carlosingles:
Thanks :+1:

I've updated the specs to use the specs from sublime-tabs and have also updated them to suit this fork.

Also I removed the case for Images since my fork is only testing for TextEditors, I'm not sure what the preferred behaviour would be for images - for me it makes more sense that images open in non-transient tabs.

carlosingles:
Thanks :+1:

I've updated the specs to use the specs from sublime-tabs and have also updated them to suit this fork.

Also I removed the case for Images since my fork is only testing for TextEditors, I'm not sure what the preferred behaviour would be for images - for me it makes more sense that images open in non-transient tabs.

roman01la:
:+1:

acontreras89:
:+1:

peterjeliot:
I think "useTransientBehavior" is much more confusing/obtuse. If I were googling for the setting, I would google "atom temporary tabs", not "atom transient behavior".

Transient, as a word, is much less accessible/recognizable than temporary. And the reason to say "tabs" despite already being in a tabs context is that it's describing the tabs. Yes, it is more specifically the tabs' behavior, but nobody would google that.

peterjeliot:
Plus, I just noticed, we didn't even agree on how to spell behavior/behaviour because of American/British differences.

jmeas:
> I think "useTransientBehavior" is much more confusing/obtuse. If I were googling for the setting, I would google "atom temporary tabs", not "atom transient behavior".

Agree. I wonder if there's some middle-ground here. Of the two options, I *much* prefer `temporaryTabs`.

Another idea would be to make this the default behavior, and then find a name for the existing behavior of the tabs. One idea is `permanent` or `persistent` are two possible words to pair with either `tabs` or `mode`. So, for instance, `permanentMode` or `persistentTabs`.

mnquintana:
I agree that `useTransientBehavior` is pretty confusing – how about calling this feature `previewTabs`? I think that gets at its purpose much more directly than `temporaryTabs`. (Why are you opening temporary tabs? To preview a file.) Plus, the `preview` moniker can be used for the corresponding features in other Atom core packages (e.g. fuzzy-finder should also be able to preview files, as should the tree-view, etc).

peterjeliot:
For comparison, the setting in Sublime Text is `preview_on_click`, which I think is just okay.

thomasjo:
> how about calling this feature previewTabs?

Great suggestion :+1: 
Perhaps with a slightly modification along the lines of `usePreviewTabs`, just to really drive it home that this is a feature flag that enables *preview tabs*, as opposed to previewing tabs or some such. I might be overthinking it though.

jmeas:
> how about calling this feature previewTabs?

I'm also a fan of this name!

> Perhaps with a slightly modification along the lines of usePreviewTabs

:+1: to that, too. I think they're both good names.

If we're in agreement that something like this is a fine name for the feature, then maybe we can merge this (if it's otherwise ready to ship) and move any fine-tweaking to the name into a separate issue. I'd hate to see something like a name tweak hold up an otherwise-good PR!

iamstarkov:
:+1: @carlosingles

peterjeliot:
:+1: for `usePreviewTabs`

carlosingles:
I've gone ahead and updated the variables to use the PreviewTabs pattern, including the flag that is set on TabViews

lib/tab-bar-view.coffee
```diff
@@ -46,7 +46,11 @@ class TabBarView extends View
     @subscriptions.add @pane.onDidRemoveItem ({item}) =>
       @removeTabForItem(item)
 
-    @subscriptions.add @pane.onDidChangeActiveItem =>
+    @subscriptions.add @pane.onDidChangeActiveItem (item) =>
+      if atom.config.get('tabs.usePreviewTabs') and item instanceof TextEditor
+        if @tab instanceof TabView and @getTabs().length > 1 and @tab.item isnt item and @tab.isPreviewTab and @tab.item instanceof TextEditor
```
kevinsawicki: Just curious, do you need the `@tab instanceof TabView` check? Won't it always be an instance of `TabView`?

Are you just trying to check for it being defined? `if @tab?` would work for that.

kevinsawicki: I don't think you need the `@tab.item instanceof TextEditor` check here either.

`@tab.isPreviewTab` will only be `true` when the item is a `TextEditor` right? Since that check is done in `TabView:: initialize`?

maletor:
This might be slightly off topic, but I've always felt that there should be support to open a file in the current tab. Preview tabs provides some support for this but most of the time Cmd-t will always open a new tab.

Would it be easy to bind something like Cmd-Shift-T to open a file in the current tab?

kevinsawicki:
> Would it be easy to bind something like Cmd-Shift-T to open a file in the current tab?

It could be made an option when selecting from the finder instead of when opening.

Like `shift` click on a file in the fuzzy finder opens the file in the current tab and `shift-enter` also does the same.

maletor:
I think this makes more sense as you propose -- to defer it to selecting a file. Maybe it's harder to implement that way? I'd be interested too see if other people prefer it one way or the other.

peterjeliot:
@maletor: I'd probably use that if it were an option.

The analogous thing in Chrome is that when you type something in the URL bar and hit `cmd`+`enter`, it opens in a new tab instead of the current one. But with that, the default behavior is to open in the current tab, and the modified behavior is to open in a new one.

Regardless, I would strongly prefer to defer it to later; I would be the type of person to only remember whether I wanted a new tab after I'd started typing the filename.

As for a workaround you can always hit `cmd`+`w`, then `cmd`+`t`.

huseyinkozan:
+1

kevinsawicki:
Awesome, thanks for this.

jmeas:
Woo! :dancer: :sparkles: :tada:

acontreras89:
Sweet
Thanks for the good work! :smiley:

thomasjo:
Awesome! :metal:

maxbrunsfeld:
A lot of people are going to love this. Nice job @ddavison and @carlosingles!

kevinsawicki:
I created a follow-on pull request to make a few tweaks, https://github.com/atom/tabs/pull/145

The main one is making this off by default to start to be conservative and allow more testing and feedback before it is enabled for everyone.

ddavison:
:D !!!!

dioxmio:
Nice work :+1:

PR #121, opened 2015-02-18
Added tab-related activationCommands

ultramiraculous:
Pulling out tab-related commands to the workspace level so commands like "Close All Tabs" are generally available from the command pane.

PR #149, opened 2015-05-28
Make commands available in the entire workspace

jazzpi:
Currently, the commands from the `tab` package are only available from the command palette if you open it after clicking on the tab bar.

This PR makes them available in the command palette regardless from where it's opened.
This makes e.g. adding a keybind for 'Close All Tabs' easy.


jazzpi:
Whoops - I only just saw #121. That one seems to be more fleshed out, so I'm closing this PR in favor of that.

PR #150, opened 2015-05-28
Delay pane activation on mousedown to fix loss of focus

nathansobo:
Force pushed an update to this pull request. This should now fix atom/atom#6662 without breaking dragging.

There's a complex interaction going on between `focus`, `mousedown`, and `dragstart` events. It seems that no matter what I try, if `document.activeElement` changes synchronously during a `mousedown` event, the subsequent `dragstart` event is never handled. There may be a hidden issue regarding our focus handling in core that would enable a more elegant fix to this, but I haven't been able to find it and would like to just fix the immediate issue for now.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -63,7 +63,7 @@ class TabBarView extends View
         false
       else if which is 1 and not target.classList.contains('close-icon')
         @pane.activateItem(tab.item)
-        @pane.activate()
+        setImmediate => @pane.activate()
         true
       else if which is 2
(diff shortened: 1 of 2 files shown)


izuzak:
@nathansobo This seems to break the ability to drag tabs (e.g. to reorder them in the tab bar, or to drag them to another windows). Are you seeing the same problem with these changes applied?

nathansobo:
@izuzak Want to give this a spin?

izuzak:
@nathansobo Works great, as far as I can tell! :zap:

Issue #142, opened 2015-05-17
Use git status colors in tab titles


grizzilus:
This is a feature request to match the color of the titles of tabs to those in the tree view based on the git status of the files. There is currently a package for this (`git-tab-status`), but I think it would be nice to have on the core tabs package to match the functionality of the `tree-view` package. (Also, I don't think `git-tab-status` is compatible with all themes. There was a point where a change in `one-dark-ui` conflicted with the methodology of that package.)

And having the `status-added` and similar classes in the core tabs package could allow theme designers to use those as well. For example, the `one-dark-ui` package has a blue bar on the left side of the active tab. This color could be changed to the color for the git status of that file. (That trivial example would obviously need some extension for active vs. inactive tabs, but the point is that this information could be integrated into themes in clever ways.)

mnquintana:
:+1: I'm definitely on board with this – I've wanted this for awhile now.

acontreras89:
Just created #152 to add this functionality.

grizzilus:
This was completed in #152 so I'll close this.

jakesankey:
This great news. I am the dev of git-tab-status, I will unpublish the package. Glad it made it into Tabs!

acontreras89:
@jakesankey :ok_hand:
thanks for your package, I used to use it until I decided to include it in the tabs core package

kankaristo:
This doesn't seem to be working for me... The file names are colored in the tree view, but not on the tabs.

The git-tab-status package is working fine, but "Enable VCS Coloring" isn't.

mnquintana:
@kankaristo It won't work if you have both enabled. You'll need to uninstall or disable git-tab-status first, then Enable VCS Coloring.

kankaristo:
I did that (first disabled, then uninstalled).

I just noticed that this doesn't seem to be working well with symlinks...

The tabs are colored fine if the project is in a normal directory, but I have a few projects where the project directory's parent directory is a symlink to an external drive, and the coloring is broken for those. The coloring works in tree-view and with git-tab-status.

mnquintana:
@kankaristo Hmmm, would you mind making a new issue for that so we can investigate further? (That's why we didn't enable it by default, just to be able to catch any lingering bugs.) Thanks!

kankaristo:
Sure. :)

acontreras89:
Point to the issue here and I'll start working on it right away :)

kankaristo:
[...]

PR #152, opened 2015-05-31
Show git status in tabs

acontreras89:
Adds git status indicators to tabs by adding the corresponding css class.

tabs_with_git_status

Addresses #142.


roman01la:
:+1:

mnquintana:
Awesome! I'll take a closer look at this later, but would you mind making this configurable? (With a setting like `enableVcsColoring`.) It's going to be configurable for the other packages (like [tree-view](https://github.com/atom/tree-view/pull/306)) so it would be great to have that here too.

acontreras89:
Working on it :ok_hand:

grizzilus:
If you create a new file from within atom, the color is updated in the tree view but not the tab. In this implementation the repo is assigned to the tab when it is initialized; the tree view package finds the repo within `updateStatus`.

lib/tab-view.coffee
```diff
@@ -1,9 +1,14 @@
 path = require 'path'
 {$} = require 'atom-space-pen-views'
+{CompositeDisposable} = require 'event-kit'
```
mnquintana: This is actually already exported from Atom so you could just do:

```coffee
{CompositeDisposable} = require 'atom'
```

and get rid of the explicit event-kit dependency.

mnquintana:
Also could you add some specs to this? Thanks!

acontreras89:
Thanks @grizzilus !
It was a little bit trickier because it had to do with the path of the file, not the repository. This also caused other inconsistencies when using the `Core: Save As` command.
I've added some logic to handle `did-save` events in cca88b20553675ec8189e8da4e76540adc27ee47. Now it should be fine :)

acontreras89:
@mnquintana thanks for the comments.
I've removed the `event-kit` dependency in 14df4cd859bdfd445ecac6bfbc7432541556147e.
I'm working now on making this feature configurable (should be ready soon). Will write some specs after that.

acontreras89:
Added specs and some minor tweaks

acontreras89:
@mnquintana I've reviewed all your comments and committed necessary changes.

I also have been working on replacing the use of `atom.project.getRepositories` (which will be removed in version 2.0) for `atom.project.repositoryForDirectory`, but I don't know how to treat one particular scenario:
The `Promise` returned by `repositoryForDirectory` should never be rejected (unless something goes really wrong). It either resolves to the repo or to `null`. 
However, when running specs, in some of them the repo has already been destroyed by the time the `Promise` is resolved. 
An error is thrown from within the `then` clause (e.g. when trying to add event listeners for the repo), giving control to the `catch` clause. What should be done in this kind of scenario? 
Note that all specs are passed despite this problem.

lib/tab-view.coffee
```diff
@@ -1,9 +1,14 @@
 path = require 'path'
 {$} = require 'atom-space-pen-views'
+{CompositeDisposable} = require 'atom'
 
 module.exports =
 class TabView extends HTMLElement
   initialize: (@item) ->
+    @path = @item.getPath?()
+
+    @subscriptions = new CompositeDisposable()
```
mnquintana: Does this assignment need to happen in `initialize`? It seems like it would be better encapsulated in `setupVcsStatus`

mnquintana: Also, since this `CompositeDisposable` is just being used for repo subscriptions, how about calling it `@repoSubscriptions`?

acontreras89: `@path` is being initialized inside `initialize` because other methods (like `updateTooltip` and `updateDataAttributes`) make use of it as well.
I'll rename `@subscriptions` (I had a mental note for it :sweat_smile:) right now.

mnquintana:
Right now, all of the VCS setup is happening even if `enableVcsColoring` is `false`, which is all unnecessary if that config option is disabled – can you update the all of the VCS setup stuff to only happen if `enableVcsColoring` is `true`? (Currently that config option is only being used to control the classes being added for the actual coloring itself.) Thanks!

acontreras89:
Sure, no problem :ok_hand:

acontreras89:
Ready @mnquintana

izuzak:
@benogle @kevinsawicki You wanted to take a look at this one, if I remember correctly.

benogle:
@acontreras89 we talked about this yesterday, and we'd like to just add data attributes to the tabs with no default styling. That way folks can add the styling they like in their styles.less, and themes can style it as well.

mnquintana:
@benogle Isn't that how this PR is working now though? It's not adding any styles, it's just using the VCS status styles defined in the active UI theme.

benogle:
You're right, there is no explicit styling in here, but it is on by default. Maybe the best option would be to default the option to `false` for now. We can default to true after it's out in the wild for a bit.

acontreras89:
@benogle changed it to have `enableVcsColoring` set to `false` by default.

lib/tab-view.coffee
```diff
@@ -68,9 +68,21 @@ class TabView extends HTMLElement
       @modifiedSubscription = dispose: =>
         @item.off?('modified-status-changed', modifiedHandler)
 
+    itemSavedHandler = (event) =>
```
benogle: Any reason this isnt inlined on line 78?

lib/tab-view.coffee
```diff
@@ -68,9 +68,21 @@ class TabView extends HTMLElement
...
+        @path = event.path
+        @setupVcsStatus() if atom.config.get 'tabs.enableVcsColoring'
+
+    if typeof @item.onDidSave is 'function'
+      @saveSubscription = @item.onDidSave(itemSavedHandler)
+
     @configSubscription = atom.config.observe 'tabs.showIcons', =>
       @updateIconVisibility()
 
+    @vcsConfigSubscription = atom.config.observe 'tabs.enableVcsColoring', (isEnabled) =>
```
benogle: Can we roll `@vcsConfigSubscription`, `@saveSubscription` etc into a composite disposable? Seems like all subscriptions that are created in init, then disposed in destroy could be under one `CompositeDisposable`

benogle:
Other than the small issues I commented on, this looks good to me.

lib/tab-view.coffee
acontreras89: Not really, just keeping the same structure as other subscriptions in the file. Wasn't really necessary since I'm not supporting old events. Changing it :art:.

lib/tab-view.coffee
acontreras89: It is true that `did-save` events might affect VCS status, but they can also require other actions (a recent example is clearing preview state).
Since I couldn't figure out a name that made sense for both `@saveSubscription` and `@vcsConfigSubscription`, I kept them apart.
On the other hand, `@repoSubscriptions` are disposed when `enableVcsColoring` is disabled, whereas `@vcsConfigSubcription` should stay put.

I can change it no problem (at least for the ones with the same lifespan) but please confirm that's the way to go :ok_hand:. 
If that's the case, should I add other subscriptions to the `CompositeDisposabe` as well? I was trying to avoid changing code that isn't related to this PR.

acontreras89:
@benogle I've replied to your comments and uploaded some changes. I have also prepared commits f0a83218218b66d0f730535126aab576da1a3b87 and f707833d963cec1b273d1e6d0ee5c79a0d19f543 to aggregate most `disposable`'s. Just let me know if you want me to merge them to this PR.

benogle:
@acontreras89 Yeah, I say we get those commits into this PR

acontreras89:
done! :+1:

mnquintana:
@benogle :ship:?

benogle:
Yep, :boat:

mnquintana:
@acontreras89 I'm super excited about this. Thanks so much for adding it :blue_heart:

acontreras89:
@mnquintana I'm glad to hear that :relieved: 
Thank to all of you for the comments! Keep up the good work :v:

thomasjo:
Awesome job! :clap:

PR #153, opened 2015-06-02
Preview tab tweaks

kevinsawicki:
Follow on to #145 

- [x] Disable preview tabs by default to now, will revisit in the future once it has been shipped for longer.
- [x] Minor formatting tweaks
- [x] Add command/context menu to keep a preview tab `tabs:keep-preview-tab`
- [x] Fix issues with clearing preview tabs across editor restarts

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -15,7 +15,7 @@ module.exports =
       default: 120
     usePreviewTabs:
       type: 'boolean'
-      default: true
+      default: false
       description: 'Tabs will only stay open if they are modified or double-clicked'
 
   activate: ->
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -90,9 +90,6 @@ class TabBarView extends View
     RendererIpc.removeListener('tab:dropped', @onDropOnOtherWindow)
     @subscriptions.dispose()
 
-  isPreviewableItem: (item) ->
-    atom.config.get('tabs.usePreviewTabs') and item instanceof TextEditor
-
   setInitialPreviewTab: ->
     activeItem = @pane.getActiveItem()
     for tab in @getTabs() when tab.isPreviewTab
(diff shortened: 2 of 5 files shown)


acontreras89:
@kevinnathan when dragging a tab from one panel to another, it becomes a temporary tab regardless of its original state. Is this the intended behavior?
I've created #161 reporting this, but we could add it to your list and close the issue if that suits you.

kevinsawicki:
> I've created #161 reporting this, but we could add it to your list and close the issue if that suits you.

A new issue is great, thanks!

franciscop:
Is there a way to disable this? I am used to the old way and definitely more used to Ctrl+W when needed than to double click.

acontreras89:
@franciscop you just have to uncheck the `Allow Pending Pane Items` in core settings.

On a side note, you can use <kbd>Ctrl-S</kbd> (or whatever your `core:save` keybinding is) to clear the pending state of the current tab. No need to double click from the start.

franciscop:
Oh awesome, I thought it was in the package `Tabs` but couldn't find it. Thank you :smile:

acontreras89:
Indeed this option *was* located under the `tabs` package when first implemented.
This has changed recently, though, as there were many issues caused by this functionality being handled exclusively by the `tabs` package.

miguelsan:
Did you test this long enough in order to enable Preview Tabs? I am looking forward to it...

50Wliu:
> Did you test this long enough in order to enable Preview Tabs? I am looking forward to it...

This is already on by default, but under a different name called "Pending Pane Items" in the Core settings.  Make sure you're on the latest Atom version, currently 1.10.2.

miguelsan:
Thank you, @50Wliu, but that I have already. What I expected is automatically closing when they lose focus. I thought that this option would allow such behaviour.

50Wliu:
> What I expected is automatically closing when they lose focus.

I believe there is an open issue for this already.

miguelsan:
Truly, #196 is it. Thanks.

Issue #157, opened 2015-06-03
Preview Tabs only work with editors


mnquintana:
**Expected:** Opening a file that doesn't open an editor (e.g. an image) will open a preview tab.
**Actual:** Opening a file that doesn't open an editor (e.g. an image) opens a regular tab.

This is because of this line: https://github.com/atom/tabs/blob/master/lib/tab-view.coffee#L8

PR #158, opened 2015-06-03
Make preview tabs work with all files

mnquintana:
Fixes https://github.com/atom/tabs/issues/157

This makes preview tabs work with any tab item that implements `::getPath()` instead of just instances of `TextEditor`.

@atom/feedback


maxbrunsfeld:
:+1: Good call.

mnquintana:
Do you think it's worth adding specs testing this with at least one filetype that won't open an editor (ie. an image)?

lee-dohm:
I think so, yes. We don't want it accidentally breaking.

mnquintana:
Alright, this should be ready for review.

lee-dohm:
:+1: Looks good to me!

kevinsawicki:
Yeah, I was deliberately defensive about this to start to prevent non-editor pane items from accidentally becoming preview tabs.

Say for example find and replace implemented a `getPath` method on their model class for some legitimate reason, it would now be a previewable item which seems less than ideal since that is something you'd want to stay open.

lee-dohm:
Should we have a `supportsTemporaryView` attribute on models then? Rather than conflating whether they have a `getPath()` method and supporting temporary mode.

mnquintana:
Hmm that's a good point, but this really should work with all files that Atom can open, instead of just text files. Another possibility is just adding more conditions to check for each possible file type that Atom can open, but that would be kinda brittle.

mnquintana:
> Should we have a supportsTemporaryView attribute on models then? Rather than conflating whether they have a getPath() method and supporting temporary mode.

Oh yeah, that's a much better idea. Maybe `supportsPreviewTabs` though instead? (Just want to try to keep the terminology consistent.)

kevinsawicki:
> Maybe supportsPreviewTabs though instead?

Maybe `isPreviewable`?

lee-dohm:
Yeah ... though #6817 suggests making a "viewer" API and I feel like "viewer" and "preview" are too close in terms maybe?

mnquintana:
@lee-dohm I was worried about that when I suggested the `preview` name for this feature, but I think they're distinct enough as long as we're consistent about it. (The only confusing thing being markdown-preview... hm.)

lee-dohm:
I'll change the name of the suggested API to "viewer" then to be more distinct.

kevinsawicki:
I'm okay leaving this using `getPath` to start and then we can then learn what packages this might impact  better and make it more opt-in in a second pass with a well-named API to be implemented by pane items.

Does that sound okay?

lee-dohm:
Agreed :+1:

lib/tab-view.coffee
```diff
@@ -5,7 +5,7 @@ path = require 'path'
 module.exports =
 class TabView extends HTMLElement
   initialize: (@item) ->
-    @isPreviewTab = atom.config.get('tabs.usePreviewTabs') and @item instanceof TextEditor
+    @isPreviewTab = atom.config.get('tabs.usePreviewTabs') and @item.getPath?
```
kevinsawicki: I'm going to update this to be `typeof @item.getPath is 'function'` just to be a little stricter.

mnquintana:
Sounds good :+1:

Sorry for jumping the gun a bit on merging this :grimacing:

Issue #61, opened 2014-06-17
Commands matching tab contextual menu (right-click on a tab)


douglascamata:
I'm missing a lot commands for the tab contextual menu actions, like "Close all", "Close unmodiffied", "Close saved", etc and I guess many other users are missing them too.

mjrk:
+1 here, too.

halleycarleton:
**+1** - I was surprised to find no default key binding for `close all tabs`.

titosrevenge:
+1 Would also like the keybinding for closing all tabs without closing the window

rubencaro:
+1. closing all tabs without closing the window a must.

lee-dohm:
There is a command for closing all tabs. And as far as I know, there are commands for all the other items in the context menu too:

<img width="613" alt="screen shot 2015-07-12 at 8 58 36 pm" src="https://cloud.githubusercontent.com/assets/1038121/8642426/cb5605c8-28d8-11e5-9dda-b2f025e3f020.png">

Closing this as fixed. If there are specific commands that are still missing, please open a single issue for each.

mryellow:
Debian KDE - Atom 0.2
![atom_close_menu](https://cloud.githubusercontent.com/assets/213575/8642723/f50034be-296b-11e5-82ed-422093d85993.png)

rubencaro:
Sorry, what we meant was to be able to define keybindings for these commands. At least that's what I meant.

lee-dohm:
You should be able to define key bindings for virtually any command in the command palette. The command name is a simple transformation from the label in the command palette. In this case:

* `tabs:close-other-tabs`
* `tabs:close-saved-tabs`
* etc

See the code here:

https://github.com/atom/tabs/blob/3b4124593b42dcdf003c0ac9a7e6c50fea34e11f/lib/tab-bar-view.coffee#L17-L23

rubencaro:
Nice!  Thanks!

mryellow:
Your competition all have a "Close All Tabs" button on the file menu. This is an expected paradigm. Missing this button leaves new users not able to find basic functionality.

PR #166, opened 2015-06-07
Register close commands on pane level

jazzpi:
Basically does the same as #121, but with working context menu commands and added specs.

This makes the commands for closing tabs available from the command palette anywhere in the `atom-workspace` scope. This makes it possible to add keybindings for things like "Close other tabs" etc. or closing all tabs on startup.

Addresses #61.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -17,6 +17,13 @@ class TabBarView extends View
     @subscriptions.add atom.commands.add atom.views.getView(@pane),
       'tabs:keep-preview-tab': => @clearPreviewTabs()
 
+    @subscriptions.add atom.commands.add 'atom-workspace',
+      'tabs:close-tab': => @closeTab()
+      'tabs:close-other-tabs': => @closeOtherTabs()
+      'tabs:close-tabs-to-right': => @closeTabsToRight()
+      'tabs:close-saved-tabs': => @closeSavedTabs()
+      'tabs:close-all-tabs': => @closeAllTabs()
+
     @subscriptions.add atom.commands.add @element,
       'tabs:close-tab': => @closeTab()
       'tabs:close-other-tabs': => @closeOtherTabs()
@@ -164,7 +171,8 @@ class TabBarView extends View
 
   closeTab: (tab) ->
     tab ?= @children('.right-clicked')[0]
-    @pane.destroyItem(tab.item)
+    tab ?= @tabForItem(@pane.getActiveItem())
+    @pane.destroyItem(tab.item) if tab?
 
   splitTab: (fn) ->
     if item = @children('.right-clicked')[0]?.item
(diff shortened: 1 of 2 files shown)


lib/tab-bar-view.coffee
```diff
@@ -17,6 +17,13 @@ class TabBarView extends View
     @subscriptions.add atom.commands.add atom.views.getView(@pane),
       'tabs:keep-preview-tab': => @clearPreviewTabs()
 
+    @subscriptions.add atom.commands.add 'atom-workspace',
+      'tabs:close-tab': => @closeTab(@getActiveTab())
```
kevinsawicki: I think it might be better to scope these to the associated pane (see line 18).

Right now if you have two split editors and you run `tabs:close-tab` from the command palette it closes the tab in both panes when I would expect it only close the tab in the pane that has focus.

jazzpi: Good catch, I did not think of splits. This would mean that closing tabs is impossible while the focus is on the tree view, but i guess that isn't really that bad.

kevinsawicki: > This would mean that closing tabs from the command palette is impossible while the focus is on the tree view

Good point, you could check `if @pane.isActive()` before calling things like `@closeTab(@getActiveTab())` to cover that case I think.

jazzpi: I don't think that's necessary since the command can only be issued when the focus is on the pane if it is only registered for the pane, or am I missing something here?

kevinsawicki: > or am I missing something here?

Sorry, I was suggesting using `if @pane.isActive()` if you went back to using the `atom-workspace` selector so that you could run the commands from the tree view.

jazzpi: That doesn't work, `@pane.isActive()` seems to return true regardless of whether the pane is focussed or not. Also, from which pane should the tab be closed if neither is focused? I think moving the scope to the pane is the best option.

kevinsawicki: > I think moving the scope to the pane is the best option.

Sounds good, want to make that update and I'll merge this in?

jazzpi: There we go, updated the specs as well.

lib/tab-bar-view.coffee
```diff
@@ -159,12 +164,16 @@ class TabBarView extends View
       @element.querySelector('.tab.active')?.classList.remove('active')
       tabView.classList.add('active')
 
+  getActiveTab: ->
+    console.log @pane.getActiveItem(), @tabForItem(@pane.getActiveItem())
```
kevinsawicki: Could you remove this line?

jazzpi: Whoops... I should remember to do a search for `console.log` before doing a PR.

kevinsawicki:
:ship: Thanks for this, it will be included in the next Atom release, 0.208

Issue #182, opened 2015-06-18
"Enable VCS Coloring" does not work with symbolic links


kankaristo:
The "Enable VCS Coloring" feature seems to break if the project directory or any of its parent directories is a symbolic link.

I have most of my (larger) projects on an external drive, because my main hard drive is a small SSD. I have symbolic links from inside my home directory to the external drive.

If I open a directory from the external drive, the tab coloring works fine. If I open the same directory through a symbolic link, the tab text color is always the same, regardless of the VCS (Git) status of the file.

The tab titles are colored correctly with the git-tab-status package, but not using "Enable VCS Coloring". The file names are also colored correctly in tree-view.

- Ubuntu 14.04.2 LTS
- Atom 0.210.0
- tabs 0.81.0

acontreras89:
I've been able to reproduce this error.
I'm already looking into it. `getRepoForPath` ([here](https://github.com/atom/tabs/blob/master/lib/tab-view.coffee#L210)) does not find the repository for the file if any of its parent folders is a symbolic link.
To fix that, `atom.project.getDirectories` should be used instead of `atom.project.getPaths` (which returns the real paths for the projects.)

.. Working on it

acontreras89:
@kankaristo can you check this issue https://github.com/atom/tree-view/issues/484 and let me know if you are able to reproduce it? 

It might be closely related to the problem we're addressing here.

acontreras89:
Created https://github.com/atom/tabs/pull/184 to fix this problem.

PR #184, opened 2015-06-18
Fix tabs not showing VCS status when parent is symlink

acontreras89:
Fixes https://github.com/atom/tabs/issues/182.
Also, replaces use of `atom.project.getRepositories`, which will be removed in version 2.0, for `atom.project.repositoryForDirectory`

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -190,9 +190,12 @@ class TabView extends HTMLElement
 
   setupVcsStatus: ->
     return unless @path?
-    repo = @repoForPath(@path)
-    @subscribeToRepo(repo)
-    @updateVcsStatus(repo)
+    @repoForPath(@path).then (repo) =>
+      @subscribeToRepo(repo)
+      @updateVcsStatus(repo)
+    .catch (err) ->
+      # We can only get here if repository is destroyed
+      # by the time the promise is resolved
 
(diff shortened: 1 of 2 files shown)


lib/tab-view.coffee
```diff
@@ -190,9 +190,13 @@ class TabView extends HTMLElement
...
-    repo = @repoForPath(@path)
-    @subscribeToRepo(repo)
-    @updateVcsStatus(repo)
+    @repoForPath(@path).then (repo) =>
+      @subscribeToRepo(repo)
+      @updateVcsStatus(repo)
+    .catch (err) ->
+      # We can only get here if repository is destroyed
+      # by the time the promise is resolved.
+      # This happens during tests
```
acontreras89: Can someone take a look at this and let me know if I should do anything in the `catch` clause?

benogle: Where does it error? Looks like both of the methods called in then check for repo existence. Can we make sure that the repo is alive in those methods?

acontreras89: Sorry, it looks like this clause wasn't even needed.

benogle:
@mnquintana you have a lot of knowledge of this. I'm happy with this.

mnquintana:
@acontreras89 Sorry for the delay - thanks so much for fixing this! :ship:

acontreras89:
@mnquintana my pleasure!

PR #191, opened 2015-07-01
Change tab from preview to permanent when press enter on file in the tree view

jwuwork:
Similar to how double-clicking on a file makes it a working file, pressing the enter key on the file should do the same.

Right now, if Preview Tabs is on and you press enter to select a file, it goes into a preview pane, which is counter-intuitive.  This commit fixes it so when you select a file by pressing enter, it treats that selection as a working file and opens it in a permanent pane.


50Wliu:
I believe this PR has been superseded by the "Pending Pane Items" feature.  Thanks for your hard work and dedication though @jwuwork!

Issue #156, opened 2015-06-03
"Use preview tabs" says tabs will stay open if double-clicked, but they don't


lee-dohm:
Using Atom v0.206.0-4941229 on Mac OS X 10.10.3.

![screen shot 2015-06-02 at 7 38 29 pm](https://cloud.githubusercontent.com/assets/1038121/7951649/0ca54bf4-095f-11e5-9aa6-84c1474b35bd.png)

### Repro Steps

1. Open Atom
1. Open Settings View
1. Navigate to Tabs package settings
1. Ensure "Use Preview Tabs" is checked
1. Close Settings View
1. Double-click any file in Tree View

**Expected:** Tab's title to not be italic (italic is the indicator that it is temporary)
**Actual:** Tab's title is italic

1. Double-click another file in Tree View

**Expected:** Second tab to open
**Actual:** New file overwrites the tab of the original file

lee-dohm:
I figured it out while trying a few more things. If you double-click the tab itself (not the file in the Tree View), it becomes permanent. So ... perhaps not a bug? I still think this will be a common misconception.

mnquintana:
This only works currently if the tab itself is double-clicked, but not if the filename is double-clicked from the tree-view. I think we'd have to tweak the behavior of clicking on files in the tree-view a bit. Currently, single-clicking opens a file but keeps focus on the tree-view, and double-clicking focuses the editor; we'd have to change it so that single-clicking opens a preview tab and focuses the editor, and double-clicking opens the file and focuses the editor.

mnquintana:
FWIW we probably should add this enhancement because it's how the feature works in Sublime and it would feel much more intuitive, I think.

lee-dohm:
> we'd have to change it so that single-clicking opens a preview tab and focuses the editor

The reason why single-clicking does not focus the editor is so that you can click a file in the Tree View and press <kbd>Backspace</kbd> to delete it.

mnquintana:
Ah alright then, so updated:

### Current behavior

- Single-clicking opens a file but keeps focus on the tree-view
- Double-clicking focuses the editor

### Proposed behavior

- Single-clicking opens a preview tab but keeps focus on the tree-view
- Double-clicking opens the file as a regular tab and focuses the editor

Does that seem reasonable?

lee-dohm:
Yes, it does :grinning:

mnquintana:
[...]

PR #195, opened 2015-07-20
Make double click in Project Find Results make tabs permanent.

harai:
Make double clicking a result of Project Find Results make the tab of its file permanent.

This makes its behavior consistent with #156 .


chayner:
:+1:

50Wliu:
This PR no longer makes sense with the introduction of Pending Pane Items in core.  Thanks for the contributions @harai!

Issue #183, opened 2015-06-18
Uncaught TypeError: Cannot read property 'dispose' of undefined


DrDanRyan:
[Enter steps to reproduce below:]

1. Open Git-Log: Show
2. Try to close tab

**Atom Version**: 0.210.0
**System**: Mac OS X 10.10.3
**Thrown From**: [tabs](https://github.com/atom/tabs) package, v0.81.0

### Stack Trace

Uncaught TypeError: Cannot read property 'dispose' of undefined

```
At /Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/composite-disposable.js:25

TypeError: Cannot read property 'dispose' of undefined
  at /Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/composite-disposable.js:25:28
  at Set.forEach (native)
  at CompositeDisposable.module.exports.CompositeDisposable.dispose (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/composite-disposable.js:24:26)
  at tabs-tab.module.exports.TabView.destroy (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tabs/lib/tab-view.js:194:14)
  at TabBarView.module.exports.TabBarView.removeTabForItem (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tabs/lib/tab-bar-view.js:364:15)
  at /Applications/Atom.app/Contents/Resources/app.asar/node_modules/tabs/lib/tab-bar-view.js:174:24
  at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:82:11)
  at Pane.module.exports.Pane.removeItem (/Applications/Atom.app/Contents/Resources/app.asar/src/pane.js:379:20)
  at Pane.module.exports.Pane.destroyItem (/Applications/Atom.app/Contents/Resources/app.asar/src/pane.js:442:16)
  at TabBarView.module.exports.TabBarView.closeTab (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tabs/lib/tab-bar-view.js:419:26)
  at Object.addElementCommands.tabs:close-tab (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tabs/lib/tab-bar-view.js:95:26)
[...]

PR #204, opened 2015-08-26
Warn when disposables outside our control are not valid

as-cii:
Fixes https://github.com/atom/tabs/issues/183.

I haven't added tests for these changes as I feel like it's fine to verify just the *happy path* (which is already covered by existing specs). Manual testing confirmed this plays nicely with git-log and the aforementioned issue.

Thoughts? :thought_balloon: I must admit I am not super in :heart: with the code, as it adds a bit of complexity to an otherwise straightforward conditional logic. What do you think?

Thanks!

/cc: @kevinsawicki @maxbrunsfeld @nathansobo

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -44,7 +44,7 @@ class TabView extends HTMLElement
       if Disposable.isDisposable(onDidChangeTitleDisposable)
         @subscriptions.add(onDidChangeTitleDisposable)
       else
-        console.warn "::onDidChangeTitle does not return a valid Disposable!"
+        console.warn "::onDidChangeTitle does not return a valid Disposable!", @item
     else if typeof @item.on is 'function'
       #TODO Remove once old events are no longer supported
       @item.on('title-changed', titleChangedHandler)
@@ -60,7 +60,7 @@ class TabView extends HTMLElement
       if Disposable.isDisposable(onDidChangeIconDisposable)
         @subscriptions.add(onDidChangeIconDisposable)
       else
-        console.warn "::onDidChangeIcon does not return a valid Disposable!"
+        console.warn "::onDidChangeIcon does not return a valid Disposable!", @item
     else if typeof @item.on is 'function'
       #TODO Remove once old events are no longer supported
       @item.on('icon-changed', iconChangedHandler)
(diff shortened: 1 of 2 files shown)


lib/tab-view.coffee
```diff
@@ -40,7 +40,11 @@ class TabView extends HTMLElement
       @updateTooltip()
 
     if typeof @item.onDidChangeTitle is 'function'
-      @subscriptions.add @item.onDidChangeTitle(titleChangedHandler)
+      onDidChangeTitleDisposable = @item.onDidChangeTitle(titleChangedHandler)
+      if Disposable.isDisposable(onDidChangeTitleDisposable)
+        @subscriptions.add(onDidChangeTitleDisposable)
+      else
+        console.warn "::onDidChangeTitle does not return a valid Disposable!"
```
kevinsawicki: Do you think it would be good to log out `@item` as well, so people know which one it was.

nathansobo: :+1:

nathansobo: Could even capture a stack trace.

maxbrunsfeld:
:+1:

nathansobo:
I think it would be pretty cheap to add a little test coverage. If it were super expensive I'd agree but maybe we should just get some coverage. I can add it if you'd like.

lib/tab-view.coffee
as-cii: Nice ideas! :sparkles: I'd prefer not to use the `@` notation, as it's strictly related to Coffeescript (and packages could be written in another language).

I'll go with something like the following:

```coffee
console.warn new Error("item::onDidChangeFoo does not return a valid Disposable").stack
```

as-cii:
> I think it would be pretty cheap to add a little test coverage. If it were super expensive I'd agree but maybe we should just get some coverage.

Sounds good, I'll add them in a while. :+1:

lib/tab-view.coffee
kevinsawicki: > Nice ideas! :sparkles: I'd prefer not to use the @ notation, though

Sorry, I meant logging out the `@item` variable directly, like as a second argument to `console.warn`

So people can inspect the object in the dev tools.

nathansobo: @as-cii I think @kevinsawicki meant to actually log out the item. It would need to be outside the string though to be useful. Honestly I think a stack trace on its might be enough. What do you think Kevin?

nathansobo: Oops, wait. A stack trace isn't going to help. Nevermind. We need to add it in core.

as-cii:
I have written some specs and updated the code to log out `@item` as well. :sparkles:

Specs are currently failing on Travis because the new Atom version (that includes [this change](https://github.com/atom/atom/commit/f602a6b726280c5d5c064abbbf950506936ecb01)) hasn't been released yet.

Thank you for your feedback, I'll wait for that commit to be :ship:ped before merging this.

as-cii:
Okay, this should be good to go now! :green_heart: 

Please, have one last :eyes: and merge at will. :bow:

nathansobo:
:dancers: Many thanks @as-cii. :dancers:

as-cii:
:heart:

PR #210, opened 2015-09-17
Add ability to create a new window by dragging a tab out or using right click menu on tab

### LINKED ISSUE #39: Pop out tabs to create new editor window

pcmccull:
Fixes #39, by adding the ability to popout a tab into a new window.

When dropping a tab it just checks if the thing that it is being dropped on does not handle the drop. So it only works if the drop effect of the item that it is hovering is "none." That way if you drop the tab onto an application that can handle the link then it doesn't create a new window. 

The window is opened using the atom command "application:new-window", which creates a new window that has one Untitled tab. The Untitled tab is then closed and the tab that the user is opening is loaded. If the item being opened in a new window has unsaved changes, they will be moved over to the new window and the tab that is left behind will be closed without warning about unsaved changes. 

Dragged and dropped windows try to open at the screen location that they were dropped. That code can be removed if you want to just have it open where ever the atom new-window command would put it.

This open tab in new window code has the same problem as the drag and drop tabs into an existing window code, where Markdown Preview and Project Find Results are blank. This seems to be a problem those packages and not something the tabs package should fix.

## Context Menu Example
contextmenunewwindow

## Drag and Drop Example 
dragtocreatewindow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,7 @@ script: 'curl -s https://raw.githubusercontent.com/atom/ci/master/build-package.
 
 git:
   depth: 10
+
+branches:
+  only:
+    - master
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -101,9 +101,11 @@ class TabBarView extends View
       false
 
     RendererIpc.on('tab:dropped', @onDropOnOtherWindow)
+    RendererIpc.on('tab:new-window-opened', @onNewWindowCreated)
 
   unsubscribe: ->
     RendererIpc.removeListener('tab:dropped', @onDropOnOtherWindow)
+    RendererIpc.removeListener('tab:new-window-opened', @onNewWindowCreated)
     @subscriptions.dispose()
 
   handleTreeViewEvents: ->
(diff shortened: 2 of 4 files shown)


tnoetzel:
Would love to see this merged in or otherwise addressed.

kradical:
Just switched over to atom (the appeal of an open source alternative to sublime is high). This is a feature that I used a lot in sublime and can't wait for it to be merged!

Thargarius:
It would be awesome if this could be done !!
And the majority of all famous code editor support this...

Benhgift:
I would really like this. Either in core or in a plugin

50Wliu:
As far as I can tell, there remains a problem when the event transfer action is not `none`, for example when what is receiving the drop is a browser window.  This is troublesome because while it can be detected if the drop event is outside the Atom window or not, it can't be cancelled.

0x15e:
Tried Atom because VS Code can't do this ... only to find Atom can't do it either. Really? Why is this not already implemented? Programmers use multiple monitors. This makes Atom a non-starter for me (same as VS Code was).

ryqiem:
This would be awesome! Please review it ❤️

PR #212, opened 2015-10-06
Make tab opening position at end option

Xcodo:
Some people (me!) like new tabs to sit at the end of the tab list, not next to the current tab.

This PR adds an option "Always open new tab at end" to the tabs package settings.

This was briefly discussed at https://discuss.atom.io/t/newly-opened-tab-order/20822

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -21,6 +21,10 @@ module.exports =
       title: "Enable VCS Coloring"
       type: 'boolean'
       default: false
+    alwaysOpenNewTabAtEnd:
+      type: 'boolean'
+      default: false
+      description: 'New tabs will be inserted at the end of the current list of tabs'
 
   activate: (state) ->
     state = [] unless Array.isArray(state)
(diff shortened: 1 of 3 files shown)


mnquintana:
Would you mind adding tests for this?

Xcodo:
Sorry that I seemed to skip over the bit about tests when reading the contributing guide.

Anyway, I've gone away and learnt what Jasmine is and added to the specs.

My tests are passing and the Travis CI build has succeeded, but my local suite fails. I'm not sure why:
atomspecfail

Xcodo:
@mnquintana This has been sitting open for a while. Will it get actioned? If it will I'll keep putting effort into resolving the conflicts with master every once in a while, but it's a bit of a pain!

simurai:
> Some people

Me too! Sometimes I like to keep a few tabs sorted. Having new tabs open in between throws off the order.

docnoe:
Looking forward to this, please merge!

50Wliu:
I believe this has already been added with the `Add New Tabs at End` option.  Really sorry for missing this PR @Xcodo :disappointed:.

PR #213, opened 2015-10-07
Loosen selector to class names to allow integration

### LINKED ISSUE #208: Loosen `[is=tree-view-file]` tree view selector requirement
The selector used for watching events in the file tree, `'.tree-view li[is=tree-view-file]'`, expects a custom element using the name "tree-view-file" but it only needs the element to have a `data-path` attribute. Switching to `.tree-view .entry.file` or allowing a selector to be configured would allow other clients to match the expected selector API without needing a custom element.

https://github.com/atom/tabs/blob/master/lib/tab-bar-view.coffee#L110

ssorallen:
`[is=tree-view-file]` enables only the custom element that registered as
`tree-view-file` to integrate with this package's preview tabs feature.
Using `.file-tree .entry.file` allows other file tree implementations to
match the selector without having to commandeer the globally-unique
"tree-view-file" custom element name.

Fixes #208

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -107,7 +107,7 @@ class TabBarView extends View
     @subscriptions.dispose()
 
   handleTreeViewEvents: ->
-    treeViewSelector = '.tree-view li[is=tree-view-file]'
+    treeViewSelector = '.tree-view .entry.file'
     clearPreviewTabForFile = ({target}) =>
       return unless @pane.isFocused()
 


ssorallen:
Every file entry in Atom's tree-view adds the class names 'file', 'entry', and 'list-item': https://github.com/atom/tree-view/blob/v0.189.0/lib/file-view.coffee#L11

thomasjo:
LGTM! Makes perfect sense. Thanks @ssorallen

PR #222, opened 2015-10-20
Drop jQuery

as-cii:
/cc: @nathansobo

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -78,7 +78,7 @@ class TabBarView extends View
     @updateActiveTab()
 
     @element.addEventListener "mousedown", @onMouseDown
-    @on 'dblclick', @onDoubleClick
+    @element.addEventListener "dblclick", @onDoubleClick
     @on 'click', '.tab .close-icon', @onClick
 
     RendererIpc.on('tab:dropped', @onDropOnOtherWindow)
@@ -370,10 +370,10 @@ class TabBarView extends View
       @pane.destroyItem(tab.item)
       preventDefault()
 
-  onDoubleClick: ({target}) =>
+  onDoubleClick: ({target, preventDefault}) =>
     if target is @element
       atom.commands.dispatch(@element, 'application:new-file')
-      false
+      preventDefault()
 
   onClick: ({target}) =>
     tab = closest(target, '.tab')
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -216,7 +216,7 @@ class TabBarView extends View
     (@paneContainer.getPanes().length > 1) or (@pane.getItems().length > 1)
 
   onDragStart: (event) =>
-    return unless event.target.classList.contains('sortable')
+    return unless event.target.matches('.sortable')
 
     event.originalEvent.dataTransfer.setData 'atom-event', 'true'
 
@@ -262,7 +262,7 @@ class TabBarView extends View
     @removePlaceholder()
 
   onDragEnd: (event) =>
-    return unless event.target.classList.contains('sortable')
+    return unless event.target.matches('.sortable')
 
     @clearDropTarget()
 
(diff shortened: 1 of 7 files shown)


lib/tab-bar-view.coffee
```diff
@@ -399,41 +420,52 @@ class TabBarView extends View
...
 
   getTabBar: (target) ->
-    target = $(target)
-    if target.is('.tab-bar') then target else target.parents('.tab-bar')
+    if target.classList.contains('.tab-bar')
+      target
+    else
+      closest(target, '.tab-bar')
+
+module.exports = document.registerElement("atom-tabs", prototype: TabBarView.prototype, extends: "ul")
```
nathansobo: Do we actually need to register a custom element? After lots of evolution in my thinking on this, I've come to think that a simple component object with a public `element` property is the best way to organize things unless we need lifecycle hooks.

nathansobo:
This is awesome, but also a pretty big change. Should we have @izuzak help us try to break it? He's good at that.

izuzak:
> Should we have @izuzak help us try to break it? He's good at that.

Happy to help :hammer: :scissors:! @as-cii ping me when you're ready -- I noticed that the specs are failing on the latest commit in this PR (and it doesn't _seem_ to be a problem with Travis :smile:).

as-cii:
Thanks, @izuzak and @nathansobo! :heart: 

Yeah, specs are failing for some reason although they pass on the latest version of Atom: something must have changed between `1.0.19` and `1.2.0-dev`. @izuzak: I'll ping you as soon as I manage to get this :green_heart: on `1.0.19`.

ssorallen:
This is great! :fireworks: 

`querySelector` and `querySelectorAll` are significantly slower than `getElementsByClassName` (https://jsperf.com/getelementsbyclassname-vs-queryselectorall/155). `getElementsByClassName` is worth considering in place of `querySelector('.single-class-name')` and `querySelectorAll('.single-classname')`.

lib/tab-bar-view.coffee
```diff
@@ -399,41 +420,52 @@ class TabBarView extends View
...
     @placeholderEl = null
 
   isPlaceholder: (element) ->
-    element.is('.placeholder')
+    element.classList.contains('placeholder')
 
   getTabBar: (target) ->
-    target = $(target)
-    if target.is('.tab-bar') then target else target.parents('.tab-bar')
+    if target.classList.contains('.tab-bar')
```
ssorallen: Extra `.` in the class name here it looks like. Should be `'tab-bar'`.

nathansobo: Thanks for the catch @ssorallen.

lib/tab-bar-view.coffee
nathansobo: After taking a look at this, I'm fine continuing with the custom element approach for now, especially since we're not using a custom tag name. Long term I'd like to just use plain elements and interact with a simple component object that points to an `element`, but for the sake of shipping this improvement that doesn't need to be now.

nathansobo:
Tests pass for me locally against master. Unclear what's going on on Travis.

nathansobo:
Okay, figured it out... The good news is that on Electron 33.x's version of Chrome, DOM fragments no longer seem to require document attachment for dispatched events to bubble. This is why the tests were passing on master. I went ahead and selectively attached DOM fragments in the tests so we can pass on older versions.

nathansobo:
Did some manual testing and fixed a couple issues I discovered. I'll run with this linked for a while. @izuzak do you want to hammer on this a bit now? @ssorallen if you want to open a PR with your optimizations against this PR, you can. Or if you don't catch it you can update your previous one against master.

izuzak:
> @izuzak do you want to hammer on this a bit now?

Yep, just did and couldn't find any problems. I'll leave it linked for a few days as well and report here if I notice any strangeness.

ssorallen:
@nathansobo I will close my PR for now and wait for this to land. The majority of the time was spent in jQuery/Sizzle, so this PR might improve the situation enough.

as-cii:
I plan to :ship: this today if everyone is :cool: with that.

Thanks everyone for the great feedback and support! :sparkles:

thomasjo:
@as-cii :shipit: :exclamation:

nathansobo:
:rocket:

as-cii:
Okay, this was just released as v0.87.0: I didn't bump the major version because it would seem like functionally-wise and from an API perspective nothing should have changed.

PR #223, opened 2015-10-21
Improve performance of `updateActiveTab`

ssorallen:
* Use native `element.getElementsByClassName`, cut out jQuery, to get
  all tabs. jQuery is nearly 100% slower than `getElementsByClassName`:
  https://jsperf.com/getelementsbyclassname-vs-queryselectorall/155
* Iterate all tabs only once when setting new active tab. `tabForItem`
  fetched and iterated all tabs once, as did `setActiveTab`, and the DOM
  was queried several times. Reduce to a single `getTabs` call (a single
  DOM query), and iterate the collection only once.

Typical slices of jQuery that appear in profiles when switching tabs are below. The stacks no longer appear with this change:

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -177,7 +177,7 @@ class TabBarView extends View
       @element.classList.remove('hidden')
 
   getTabs: ->
-    @children('.tab').toArray()
+    Array.prototype.slice.call(@element.getElementsByClassName('tab'))
 
   tabAtIndex: (index) ->
     @children(".tab:eq(#{index})")[0]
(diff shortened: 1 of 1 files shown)


ssorallen:
Not sure why the test can't find atom-space-pen-views?

`[590:1021/142753:INFO:CONSOLE(27)] "Error: Cannot find module 'atom-space-pen-views'`

ssorallen:
Closing in favor of #222 for now. jQuery + Sizzle were the time sinks here, and jQuery is removed in #222. I'll revisit after that PR is merged to see if more changes would still be helpful.

PR #230, opened 2015-11-01
Fix bad `this` binding

### LINKED ISSUE #229: Uncaught TypeError: Cannot read property 'id' of undefined
1. Have settings open
2. Make new window
3. Drag settings tab to new window
(edit: tested with a non-settings pane and it still threw this error)

**Atom Version**: 1.2.0-beta0
**System**: Microsoft Windows 10 Pro
**Thrown From**: [tabs](https://github.com/atom/tabs) package, v0.87.0

### Stack Trace

Uncaught TypeError: Cannot read property 'id' of undefined

```
At ...\AppData\Local\atom\app-1.2.0-beta0\resources\app.asar\node_modules\tabs\lib\tab-bar-view.js:586

TypeError: Cannot read property 'id' of undefined
    at EventEmitter.TabBarView.onDropOnOtherWindow (...\AppData\Local\atom\app-1.2.0-beta0\resources\app.asar\node_modules\tabs\lib\tab-bar-view.js:586:20)
    at emitTwo (events.js:92:20)
    at EventEmitter.emit (events.js:172:7)
```

### Commands

```
     -1:08.8.0 core:undo (atom-text-editor.editor.is-focused)
     -1:07.8.0 core:backspace (atom-text-editor.editor.is-focused)
     -1:07.5.0 core:save (atom-text-editor.editor.is-focused)
     -0:57.9.0 core:copy (atom-text-editor.editor.is-focused)
     -0:57.6.0 core:paste (atom-text-editor.editor.is-focused)
  2x -0:56.7.0 editor:newline (atom-text-editor.editor.is-focused)
     -0:56.4.0 core:paste (atom-text-editor.editor.is-focused)
  7x -0:56 core:backspace (atom-text-editor.editor.is-focused)
     -0:54.2.0 editor:newline (atom-text-editor.editor.is-focused)
     -0:53.8.0 core:save (atom-text-editor.editor.is-focused)
     -0:51.2.0 core:cut (atom-text-editor.editor.is-focused)
     -0:50.9.0 core:save (atom-text-editor.editor.is-focused)
     -0:31.6.0 tabs:split-up (li.tab.sortable.active.right-clicked)
     -0:27.5.0 pane:reopen-closed-item (ul.list-inline.tab-bar.inset-panel)
     -0:25.8.0 tabs:close-tab (li.tab.sortable.active.right-clicked)
     -0:21.6.0 pane:close (li.tab.sortable.active.right-clicked)
```

### Config

```json
{
  "core": {}
}
```

### Installed Packages

```coffee
# User
autocomplete-python, v0.9.4
python-isort, v0.0.7
python-tools, v0.6.5

# Dev
No dev packages
```

as-cii:
Fixes #229.

With #222 we dropped jQuery and used a custom element to encapsulate the behavior of a `TabBarView`. When instantiating a custom element, though, any custom constructor that's set on the prototype won't be called. This means that, if we're using Coffeescript's fat arrow facility to define a method on the object's prototype, this is not going to work because of how this feature is implemented in the language. e.g.

```coffee
class Foo
  something: => 1
```

It will be translated to something like the following:

```js
class A {
  constructor () {
    this.something = this.something.bind(this)
  }
  something () {
    return 1
  }
}
```

Since the constructor cannot be executed, `this` will be set to a wrong value. We were making use of this functionality in `::onDropOnOtherWindow`, which started to fail as soon as started using custom elements.

I think this is yet another signal that we need to move away from custom elements, but rewriting the whole object right now to not use them seems a bit riskier than simply fixing #229.

/cc: @izuzak @nathansobo

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -217,7 +217,7 @@ class TabBarView extends HTMLElement
   shouldAllowDrag: ->
     (@paneContainer.getPanes().length > 1) or (@pane.getItems().length > 1)
 
-  onDragStart: (event) =>
+  onDragStart: (event) ->
     return unless matches(event.target, '.sortable')
 
     event.dataTransfer.setData 'atom-event', 'true'
@@ -260,15 +260,15 @@ class TabBarView extends HTMLElement
     catch error
       false
 
-  onDragLeave: (event) =>
+  onDragLeave: (event) ->
     @removePlaceholder()
 
-  onDragEnd: (event) =>
+  onDragEnd: (event) ->
     return unless matches(event.target, '.sortable')
 
     @clearDropTarget()
 
-  onDragOver: (event) =>
+  onDragOver: (event) ->
     unless event.dataTransfer.getData('atom-event') is 'true'
       event.preventDefault()
       event.stopPropagation()
@@ -297,7 +297,7 @@ class TabBarView extends HTMLElement
         else
           element.parentElement.appendChild(placeholder)
 
-  onDropOnOtherWindow: (fromPaneId, fromItemIndex) =>
+  onDropOnOtherWindow: (fromPaneId, fromItemIndex) ->
     if @pane.id is fromPaneId
       if itemToRemove = @pane.getItems()[fromItemIndex]
         @pane.destroyItem(itemToRemove)
@@ -311,7 +311,7 @@ class TabBarView extends HTMLElement
     @removeDropTargetClasses()
     @removePlaceholder()
 
-  onDrop: (event) =>
+  onDrop: (event) ->
     event.preventDefault()
 
     return unless event.dataTransfer.getData('atom-event') is 'true'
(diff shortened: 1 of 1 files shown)


izuzak:
Thanks for looking into this, @as-cii, and for the quick fix :zap: This fixes #229 for me. :+1:

lib/tab-bar-view.coffee
```diff
@@ -82,10 +82,10 @@ class TabBarView extends HTMLElement
     @addEventListener "dblclick", @onDoubleClick
     @addEventListener "click", @onClick
 
-    RendererIpc.on('tab:dropped', @onDropOnOtherWindow)
+    RendererIpc.on('tab:dropped', => @onDropOnOtherWindow(arguments...))
```
nathansobo: What do you think about just passing in `@onDropOnOtherWindow.bind(this)`

nathansobo:
Minor stylistic comment. Feel free to merge after addressing it.

nathansobo:
:+1: Thanks!

PR #231, opened 2015-11-02
Listen to path changed event in addition to title changed

### LINKED ISSUE #189: Incorrect file name displayed in tooltip
The tab has a label that displays the file name. If you hover the cursor over that label a balloon displays the full path to the file. At times the balloon displays the wrong path, as in the following example.
1. Save "file 1" to "path 1".
2. Hover over the label.
3. The correct path of file 1 at path 1 is displayed.
4. Save As... the same "file 1" to "path 2", using the same file name. Only the path is different.
5. Hover over file name label. The displayed path is file 1 at path 1. 
6. Modify contents of file and save it.
7. Hover over the label. The displayed path is still file 1 at path 1.
8. The file that was modified, however, is file 1 at path 2!

A variation. If you change the file name at step 4, where you "Save As..." to path 2, then the label in the tab will display the new file name, but if you hover over the label, then the balloon will display the old file name with the old path.

In short, once set, information displayed in the file name and path information balloon doesn't change, even under circumstances where the label correctly displays the new file name. Moreover, regardless what path is displayed in the balloon, modifications will always be saved to the latest file name and path combination for that tab.

Hope this helps.

hgGeorg:
This should fix #189.

Tab-View currently listens to the title changed event and calls various methods to update its data/view.
The problem is that the path is never set after a rename or save-as operation so those methods use out of date data.
This PR should fix this issue by listening to path changed event instead and setting `@path` correctly.

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -33,22 +33,22 @@ class TabView extends HTMLElement
       @addEventListener 'dblclick', => @clearPreview()
 
   handleEvents: ->
-    titleChangedHandler = =>
+    pathChangedHandler = (@path) =>
       @updateDataAttributes()
       @updateTitle()
       @updateTooltip()
 
-    if typeof @item.onDidChangeTitle is 'function'
-      onDidChangeTitleDisposable = @item.onDidChangeTitle(titleChangedHandler)
-      if Disposable.isDisposable(onDidChangeTitleDisposable)
-        @subscriptions.add(onDidChangeTitleDisposable)
+    if typeof @item.onDidChangePath is 'function'
+      onDidChangePathDisposable = @item.onDidChangePath(pathChangedHandler)
+      if Disposable.isDisposable(onDidChangePathDisposable)
+        @subscriptions.add(onDidChangePathDisposable)
       else
-        console.warn "::onDidChangeTitle does not return a valid Disposable!", @item
+        console.warn "::onDidChangePath does not return a valid Disposable!", @item
     else if typeof @item.on is 'function'
       #TODO Remove once old events are no longer supported
-      @item.on('title-changed', titleChangedHandler)
+      @item.on('path-changed', pathChangedHandler)
       @subscriptions.add dispose: =>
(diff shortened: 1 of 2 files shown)


lib/tab-view.coffee
```diff
@@ -50,6 +48,23 @@ class TabView extends HTMLElement
...
+      @updateTooltip()
+
+    if typeof @item.onDidChangePath is 'function'
+      onDidChangePathDisposable = @item.onDidChangePath(pathChangedHandler)
+      if Disposable.isDisposable(onDidChangePathDisposable)
+        @subscriptions.add(onDidChangePathDisposable)
+      else
+        console.warn "::onDidChangePath does not return a valid Disposable!", @item
+    else if typeof @item.on is 'function'
+      #TODO Remove once old events are no longer supported
```
thedaniel: We don't need this conditional anymore, it's OK to just use `onDidChangePath`

thedaniel:
Thanks for this - sorry to leave it for so long. I left one comment, if you make that change this is good to go. I can also pull in your changes and update it myself if necessary.

Alhadis:
Paging @hgGeorg

lib/tab-view.coffee
hgGeorg: You mean the typeof condition right?
This is still needed?
``` javascript
onDidChangePathDisposable = @item.onDidChangePath(pathChangedHandler)
if Disposable.isDisposable(onDidChangePathDisposable)
  @subscriptions.add(onDidChangePathDisposable)
else
  console.warn "::onDidChangePath does not return a valid Disposable!", @item
```

I'm a little confused because those are everywhere.

Alhadis: Yeah, he's probably referring to `if typeof @item.onDidChangePath is function`.

thedaniel: ah, in that case, never mind. i was just looking at the diff, now that I look at the whole package in context it's fine, those can be removed by someone else in another PR.

Alhadis: So, this can be merged? =)

PR #238, opened 2015-11-19
Use new way of requiring Electron modules

50Wliu:
Refs atom/atom#9627

~~I'm not sure how to make this backwards-compatible.  I assume all it takes is a simple if block (`if x require('ipc-renderer') else require('ipc')`) but I'm not sure what the `x` would be.~~

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -1,4 +1,5 @@
 BrowserWindow = null # Defer require until actually used
+# TODO: Remove the catch once Electron 0.35.0 is bundled in Atom
 try require {ipcRenderer} = 'electron' catch then ipcRenderer = require 'ipc'
 
 {matches, closest, indexOf} = require './html-helpers'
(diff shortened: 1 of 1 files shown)


as-cii:
I think the best way we could possibly handle this is through a `try/catch` block. Basically, we try to require `ipc-renderer` and, if that fails, we go for `ipc`.

```coffee
ipc =
  try
    require 'ipc-renderer'
  catch
    require 'ipc'
```

Or, on a single line, if that looks better:

```coffee
ipc = try require('ipc-renderer') catch then require('ipc')
```

lib/tab-bar-view.coffee
```diff
@@ -1,5 +1,5 @@
 BrowserWindow = null # Defer require until actually used
-RendererIpc = require 'ipc'
+ipc = try require 'ipc-renderer' catch then require 'ipc'
```
steelbrain: I think we should drop support for previous versions because the only place where this will be included will be bumping the version

50Wliu: It was to make specs pass.  I'll remove it once a shipped version of Atom contains this.

PR #246, opened 2015-12-24
Style tab for pending item

kuychaco:
Single-clicking opens a file in pending state and double-clicking opens a file regularly (in a non-pending state).

See corresponding changes in [atom/atom/pull/10178](https://github.com/atom/atom/pull/10178)
Addresses atom/atom/issues/9165


kuychaco:
Uncaught Error during tests (present on master before changes)
screen shot 2016-01-05 at 7 08 05 pm

Issue #190, opened 2015-06-30
Cannot drag inactive tabs without focusing them first


matthewmorek:
You have to make a tab/pane active first, before you are able to click-drag it, to reorder, or move across to another pane.
#### Steps to reproduce:
1. Open up a file in Atom.
2. Split the panes in whichever way you want.
3. Click on either pane/tab to make it active/bring it to focus.
4. Try click-dragging the other file into the opposite pane.

infogulch:
This is still a problem.

More specifically, it manifests when you attempt to drag a tab that's *not in the currently active pane*.

This also negatively affects packages that rely on `tabs` for their functionality, like morassman/tabasco#6 which allows you to split panes by overloading dragging tabs.

infogulch:
After a tiny bit of investigation I found that the `onDragStart` event is never fired on tabs that are in an inactive pane, for any event listener.

I can use the debugger but I'm not a web dev and I have no idea where to proceed from here.

50Wliu:
This seems to have been resolved - I can no longer reproduce.

PR #252, opened 2016-01-27
Drag and drop layout management

MikeInnes:
layout

This kind of UI is really handy for those who are less keyboard-savvy, especially if you have a bunch of plots / workspace panes / consoles etc. open and don't want to remember the right sequence of commands needed to organise things :)

I'm still fairly new to CoffeeScript / Atom development so feedback is definitely appreciated. The global `@lastSplit` approach is a bit weird, but seems to be necessary as the `dragend` event gives different `clientX` and `clientY` coordinates which throws things off. I've aimed for an overlay styling that works well with both light and dark syntax / ui themes, but other suggestions are appreciated there as well.

### DIFF
--- a/lib/layout.coffee
+++ b/lib/layout.coffee
@@ -10,6 +10,8 @@ module.exports =
   deactivate: ->
     @view.parentElement?.removeChild @view
 
+  test: {}
+
   drag: (e) ->
     @lastCoords = e
     pane = @getPaneAt e
@@ -44,10 +46,10 @@ module.exports =
     closest document.elementFromPoint(clientX, clientY), selector
 
   getItemViewAt: (coords) ->
-    @getElement coords, '.item-views'
+    @test.itemView or @getElement coords, '.item-views'
 
   getPaneAt: (coords) ->
-    @getElement(@lastCoords, 'atom-pane')?.getModel()
+    @test.pane or @getElement(@lastCoords, 'atom-pane')?.getModel()
 
   paneForTab: (tab) ->
     tab.parentElement.pane
@@ -88,7 +90,7 @@ module.exports =
 
   updateView: (pane, coords) ->
     @view.classList.add 'visible'
-    rect = pane.getBoundingClientRect()
+    rect = @test.rect or pane.getBoundingClientRect()
     split = if coords then @splitType @normalizeCoords rect, coords
     @updateViewBounds @innerBounds rect, @boundsForSplit split
     split
--- a/lib/layout.coffee
+++ b/lib/layout.coffee
@@ -4,8 +4,8 @@ module.exports =
 
   activate: ->
     @view = document.createElement 'div'
-    @view.classList.add 'layout-overlay'
     document.body.appendChild @view
+    @view.classList.add 'tabs-layout-overlay'
 
   deactivate: ->
     @view.parentElement?.removeChild @view
--- a/lib/layout.coffee
+++ b/lib/layout.coffee
@@ -1,0 +1,97 @@
+{closest, indexOf, matches} = require './html-helpers'
+
+module.exports =
+
+  activate: ->
+    @view = document.createElement 'div'
+    @view.classList.add 'layout-overlay'
+    document.body.appendChild @view
+
+  deactivate: ->
+    @view.parentElement?.removeChild @view
+
+  drag: (e) ->
+    @lastCoords = e
+    pane = @getPaneAt e
+    itemView = @getItemViewAt e
+    if pane? and itemView?
+      coords = if not (@isOnlyTabInPane(pane, e.target) or pane.getItems().length is 0)
+        [e.clientX, e.clientY]
+      @lastSplit = @updateView itemView, coords
+    else
+      @disableView()
+
+  end: (e) ->
+    @disableView()
+    return unless @getItemViewAt @lastCoords
+    target = @getPaneAt @lastCoords
+    return unless target?
+    toPane = switch @lastSplit
+      when 'left'  then target.splitLeft()
+      when 'right' then target.splitRight()
+      when 'up'    then target.splitUp()
+      when 'down'  then target.splitDown()
+    tab = e.target
+    toPane ?= target
+    fromPane = @paneForTab tab
+    return if toPane is fromPane
+    item = @itemForTab tab
+    fromPane.moveItemToPane item, toPane
+    toPane.activateItem item
+    toPane.activate()
+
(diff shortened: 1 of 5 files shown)


lee-dohm:
Could you add specs for this so that we are sure that we don't break it in future changes?

one-more-minute:
Sure thing, I've added a couple of tests and will finish them off tomorrow – let me know if what's there so far looks ok.

lee-dohm:
:+1: Looks really good so far! Please @-mention me when you've got things fixed up so I can review it again. Thanks again for the hard work!

one-more-minute:
@lee-dohm Glad I can be helpful – and thank you for the review! I've finished off the last couple tests.

one-more-minute:
Just bumping this, would be great to get a review.

By the way, for the testing I've had to inject test values into the layout module in place of the usual `clientWidth`/`clientHeight`, but it looks like I might (?) be able to avoid that by attaching the workspace to the dom for the tests. I haven't played around with that, though, so I don't know if it will actually work; let me know if it's something I should do.

benogle:
Thanks, it's in our review queue. Our review queue, however, is long, so it might take a bit to get to.

one-more-minute:
Ok great, well it's good to know it's in the pipeline, so thanks for the update!

jerone:
I'm not a keyboard guy and applaud any UI improvements, but this is super awesome! :+1: 

I've played with it and works great. Did notice that it only works when the tab is focused, but that could be an original tabs issue. 

Was wondering about your cursor in the screengrab... I'm on Windows and there it shows just the drag cursor (cursor drag move).

lib/layout.coffee
```diff
@@ -0,0 +1,99 @@
...
+    [x, y, w, h] = switch split
+      when 'left'   then [0,   0,   0.5, 1  ]
+      when 'right'  then [0.5, 0,   0.5, 1  ]
+      when 'up'     then [0,   0,   1,   0.5]
+      when 'down'   then [0,   0.5, 1,   0.5]
+      else               [0,   0,   1,   1  ]
+
+  innerBounds: ({left, top, width, height}, [x, y, w, h]) ->
+    left += x*width; top += y*height
+    width *= w; height *= h
```
jerone: Intentional use of semicolon?

one-more-minute: Yup, I just found that laying the statements out in a rectangle made it more intuitive for me. I'll be happy to defer to the maintainers if people don't like the style, of course.

one-more-minute:
@jerone thanks for the feedback, and glad you like it! I haven't (intentionally) changed the cursor or focus behaviour, but did also find it odd that it only works with a focused tab. Could look into changing that if it seems like a good fix.

lib/layout.coffee
mnquintana: Yeah, it would be great to get this all on separate lines :smile: Thanks!

one-more-minute: Done!

one-more-minute:
There's something strange going on with the tests – they passed fine before but now fail despite no code changes (the empty commit `foo` demonstrates this). Seems to be related to pending-tab code. Let me know what, if anything, I should do about that.

lee-dohm:
The production version of Atom changed from v1.4.3 to v1.5.0 yesterday which included some pending tabs changes, I believe. This PR will need to be updated to work with that feature. (Fortunately, it is only one failing test :grinning:)

one-more-minute:
Looks like it's an issue with the [master branch of this package](https://travis-ci.org/atom/tabs), rather than being specific to this PR.

MikeInnes:
Have rebased over the now-working master and the green light is back on.

SebastianSzturo:
This is really amazing! Great work @MikeInnes :tada: :heart:

alexandernst:
Can't wait to see this merged!! :+1:

infogulch:
@MikeInnes mentioned a problem, quoted below. Issue #190 is tracking this, I just wanted to link them.

> changed the cursor or focus behaviour, but did also find it odd **that it only works with a focused pane**. Could look into changing that if it seems like a good fix.

MikeInnes:
Just bumping this again, I don't want to be impatient but it has been nearly four months since I opened the PR.

simurai:
👍  on adding this. It feels pretty intuitive and quick without having to think what command to use.

lee-dohm:
@MikeInnes Would you mind fixing the merge conflicts? Once you do, I'll take a look at getting this merged.

lib/layout.coffee
```diff
@@ -0,0 +1,101 @@
+{closest, indexOf, matches} = require './html-helpers'
+
+module.exports =
+
+  activate: ->
+    @view = document.createElement 'div'
+    @view.classList.add 'layout-overlay'
+    document.body.appendChild @view
```
simurai: Not too critical, but maybe something to consider. Move the `layout-overlay` element inside `atom-pane-container.panes`?

screen shot 2016-05-19 at 9 44 43 am

Just so the DOM root stays a bit more organized. And/or rename it to `tabs-layout-overlay`. Then it's more clear that this element is part of the `tabs` package and shouldn't be used for something else.

MikeInnes: Thanks for the tips, will go ahead and do both!

MikeInnes:
@lee-dohm will do, thanks a lot for taking a look.

lib/layout.coffee
MikeInnes: Ok so I did the rename and played around with adding the overlay to `atom-pane-container.panes`. Doing that complicates the layout logic a little (which should be fixable) but unfortunately also seems to confuse Atom's pane logic, resulting in empty panes appearing and such.

Instead I added the overlay to atom-workspace which doesn't have either of those issues but I think solves the issue of keeping the root node clean – hope that's ok.

lee-dohm:
Thanks so much for all your hard work on this @MikeInnes!

paulcbetts:
This is Right and Good, :clap: :clap: : clap:

MikeInnes:
Awesome, very happy to see this in – thank you @lee-dohm!

morassman:
Amazing! I made my own plugin for this behavior, [tabasco](https://atom.io/packages/tabasco), but I'll rather use the official implementation once it's part of Atom. Very nice!

Chrisedmo:
Damn. Nice. 👊🏼

bronson:
This is one of my favorite features of 2016.  Love using it.

Any chance of making it work when dropping the tab in a new window?  https://github.com/atom/atom/issues/14273

PR #262, opened 2016-02-15
Add the abitlity to pin tabs

ArnaudRinquin:
pinned

This adds 3 commands:

* `tabs:pin-tab`
* `tabs:unpin-tab`
* `tabs:toggle-pin-tab`

This adds the `Toggle pin tab` command to both palette commands and as
`File` sub menu.

This adds the `Pin Tab` or `Unpin Tab` context menu to tabs (depending
on current status)

Pinned tabs titles are bolden and a `lock` icon is shown (only if `showIcon` is true, overrides existing icon)

Pinned tabs are **not** closed on following events:

* close-all-tabs
* close-other-tabs
* close-tabs-to-right
* close-saved-tabs

_Note:_ pinned tabs are still closed on the regular `close-tab` command

PR comes with updated tests and new ones.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -20,6 +20,7 @@ class TabBarView extends HTMLElement
     @subscriptions.add atom.commands.add atom.views.getView(@pane),
       'tabs:keep-pending-tab': => @terminatePendingStates()
       'tabs:close-tab': => @closeTab(@getActiveTab())
+      'tabs:toggle-pin-tab': => @togglePinTab(@getActiveTab())
       'tabs:close-other-tabs': => @closeOtherTabs(@getActiveTab())
       'tabs:close-tabs-to-right': => @closeTabsToRight(@getActiveTab())
       'tabs:close-saved-tabs': => @closeSavedTabs()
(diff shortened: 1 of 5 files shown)


ArnaudRinquin:
Note that the tests are failing not because of my changes but were already failing on current `master`.

jerone:
@ArnaudRinquin Did you have a look at [pinned-tabs](https://atom.io/packages/pinned-tabs) package? I think it has some nice features that could improve this PR.

ArnaudRinquin:
@jerone The `pinned-tabs` package provides a cosmetic only (show icon only, pull to the left) implementation vs. I am proposing an implementation that brings something more logical: preventing tabs from being closed on commands targeting many tabs.

These _cosmetic_ feature could be ported to the core `tabs` package as well or left optional through the package, I don't mind, but the _logical_ part absolutely needs to be part of the code package.

ArnaudRinquin:
I updated the behaviour so the `close-icon` is replaced by a lock instead of the `file-icon`. The `close-icon` is restored when tab is hovered, as shown here:

pinned-tabs

ArnaudRinquin:
@simurai, do you have any feedback regarding the icon behaviour? @abe33 made some suggestions regarding other themes:

* make sure the lock icon is always shown, including on non-active windows (Atom Material)
* handle unsaved files icon as well

Anything else?

lee-dohm:
I apologize that it has taken a while to get back to you on this @ArnaudRinquin. Having the ability to pin tabs is not something that we're interested in having or maintaining in our tabs package. Thank you very much for all the effort you went into developing and showcasing this functionality.

ArnaudRinquin:
No worries, I'll try to get it under the form of a more independent plugin maybe.

PR #263, opened 2016-02-15
Close tab on OSX when corresponding file is moved to trash

### LINKED ISSUE #160: Tab of deleted file should be closed
When you delete a file, if it was open in a tab I think the tab should be closed.

What happens now, is that after deleting a file from within tree view, the tab remains open with the file being in trash. Then when you want to close the window, you get a confirmation screen asking you if you want to save the file or not. This is pretty annoying and seems redundant. Deleting a file already comes with it's own confirmation dialog.

If this is an issue for atom/tree-view let me know and I'll re-post there.

anderoonies:
Fixes #160 
Closes all tabs corresponding to a given file if it's moved to trash.

To test:

1. Create a new file
2. (optional) Split into an additional pane
3. Right click on the file in the tree and Delete->Move to Trash

Questions:

- where best to check for platform
- most rigorous way to check path

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -59,6 +59,7 @@ class TabBarView extends HTMLElement
 
     @subscriptions.add @pane.onDidAddItem ({item, index}) =>
       @addTabForItem(item, index)
+      @subscribeToTabPath(item)
 
     @subscriptions.add @pane.onDidMoveItem ({item, newIndex}) =>
       @moveItemTabToIndex(item, newIndex)
@@ -85,6 +86,12 @@ class TabBarView extends HTMLElement
   unsubscribe: ->
     ipcRenderer.removeListener('tab:dropped', @onDropOnOtherWindow)
     @subscriptions.dispose()
+
+  subscribeToTabPath: (item) ->
+    if typeof item.buffer?.onDidChangePath is 'function'
+      @subscriptions.add item.buffer?.onDidChangePath (path) =>
+        if process.platform is 'darwin' and /.Trash/.test(path)
+          @closeTab(@tabForItem(item))
 
   terminatePendingStates: ->
     tab.terminatePendingState?() for tab in @getTabs()


lib/tab-bar-view.coffee
```diff
@@ -86,6 +87,12 @@ class TabBarView extends HTMLElement
     ipcRenderer.removeListener('tab:dropped', @onDropOnOtherWindow)
     @subscriptions.dispose()
 
+  subscribeToTabPath: (item) ->
+    if typeof item.buffer?.onDidChangePath is 'function'
+      @subscriptions.add item.buffer?.onDidChangePath (path) =>
+        if process.platform is 'darwin' and /.Trash/.test(path)
```
acontreras89: Shouldn't the regular expression be `/^\.Trash$/` instead?

anderoonies: absolutely, i need to escape the period my mistake.
`path`, though, is the full path containing the file, so something like:
`/Users/me/.Trash/test.txt`
a more rigorous regex might be `/\/\.Trash\//`

acontreras89:
Thanks for writing back.
Please do keep an eye on the discussion in atom/tabs#160, as this might not be the best approach to solve this issue. If we simply remove the tab, other packages will still behave incorrectly after a file has been moved to the trash.

anderoonies:
100% in agreement, i [proposed](https://github.com/atom/tabs/issues/160#issuecomment-184031903) handling this in `node-pathwatcher` rather than in tabs but wanted to propose an alternative
thanks for the review

acontreras89:
https://github.com/atom/tabs/issues/160 is now fixed and this PR doesn't make sense anymore. Can we close it? :)

anderoonies:
@acontreras89 completely forgot about this after fixing pathwatcher. sorry :bow:

acontreras89:
@anderoonies no problem :ok_hand: thanks!

Issue #270, opened 2016-02-26
Duplicate `Copy Full Path` and `Copy Project Path` from tree-view


leesdolphin:
Add 2 items to the tabs context menu to copy the file and project path. The implementation in atom/tree-view is [here](https://github.com/atom/tree-view/blob/master/lib/tree-view.coffee#L318)

leesdolphin:
I'll try and create a pull request tonight

a7madgamal:
+10

PR #272, opened 2016-02-28
Add Copy Project/Full Path menu items.

software-opal:
Implements atom/tabs#270

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -44,6 +44,8 @@ class TabBarView extends HTMLElement
       'tabs:split-down': => @splitTab('splitDown')
       'tabs:split-left': => @splitTab('splitLeft')
       'tabs:split-right': => @splitTab('splitRight')
+      'tabs:copy-full-path': => @copyTabPath(false)
+      'tabs:copy-project-path': => @copyTabPath(true)
 
     @addEventListener "dragstart", @onDragStart
     @addEventListener "dragend", @onDragEnd
@@ -172,6 +174,11 @@ class TabBarView extends HTMLElement
 
   closeAllTabs: ->
     @closeTab(tab) for tab in @getTabs()
+
+  copyTabPath: (relativePath = false) ->
+    tab = @querySelector('.right-clicked')
+    ## Provided in the atom/text-editor module
+    tab.item.copyPathToClipboard?(relativePath)
 
   getWindowId: ->
     @windowId ?= atom.getCurrentWindow().id
--- a/menus/tabs.cson
+++ b/menus/tabs.cson
@@ -17,6 +17,11 @@ menu: [
 
     {type: 'separator'}
 
+    {label: 'Copy Full Path', command: 'tabs:copy-full-path'}
+    {label: 'Copy Project Path', command: 'tabs:copy-project-path'}
+
+    {type: 'separator'}
+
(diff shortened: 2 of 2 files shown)


lib/tab-bar-view.coffee
```diff
@@ -173,6 +175,11 @@ class TabBarView extends HTMLElement
   closeAllTabs: ->
     @closeTab(tab) for tab in @getTabs()
 
+  copyTabPath: (relativePath = false) ->
+    tab = @querySelector('.right-clicked')
+    ## Provided in the atom/text-editor module
+    tab.item.copyPathToClipboard?(relativePath)
```
nathansobo: Rather than calling `copyPathToClipboard` on the pane item, we should call `getURI` on the pane item and perform similar logic that occurs in `TextEditor.prototype.copyPathToClipboard` here. That will allow this command to work with all items, since they're required to implement that method. For a normal text editor, it returns its path as its URI, so this will work in the common case.

lib/tab-bar-view.coffee
```diff
@@ -44,6 +44,8 @@ class TabBarView extends HTMLElement
       'tabs:split-down': => @splitTab('splitDown')
       'tabs:split-left': => @splitTab('splitLeft')
       'tabs:split-right': => @splitTab('splitRight')
+      'tabs:copy-full-path': => @copyTabPath(false)
+      'tabs:copy-project-path': => @copyTabPath(true)
```
nathansobo: What about `tabs:project-relative-path`.

nathansobo: Come to think of it, it should probably be `tabs:copy-project-relative-uri`

lib/tab-bar-view.coffee
```diff
@@ -44,6 +44,8 @@ class TabBarView extends HTMLElement
       'tabs:split-down': => @splitTab('splitDown')
       'tabs:split-left': => @splitTab('splitLeft')
       'tabs:split-right': => @splitTab('splitRight')
+      'tabs:copy-full-path': => @copyTabPath(false)
```
nathansobo: I think this could be called `tabs:copy-uri`. With paths there's some ambiguity about absolute vs relative but with URIs I think absolute is assumed.

nathansobo:
I'm open to this but left some comments about my concerns with generality. Relying on `copyPathToClipboard` seems less than ideal when all items have a URI.

Using the URI rather than the path does open up some weirdness for the UI, since users might expect the command to be about paths (the common case). This is a tough balance between wanting to be general but also familiar in the common case. Curious as to your thoughts on it.

adamreisnz:
@leesdolphin are you able to address @nathansobo concerns?
I would love to see this PR get merged, as I just wanted to open an issue to implement this myself.

leesdolphin:
I'd completely forgotten about this. I will review them in the morning.

50Wliu:
@nathansobo I want to pick this PR back up, but I have a few questions. `getURI` is marked as an optional method for workspace items. Additionally, since there already exists [`editor:copy-[project]-path`](https://github.com/atom/atom/blob/11bd210ee3f2b5a570d27d63435c0024ed1f5e50/src/register-default-commands.coffee#L277-L278) commands, would it make more sense to use `tab.item.getPath()`, and also name the commands `tab:copy-[project]-path` for consistency with the editor ones?

PR #273, opened 2016-02-28
Allow config to add new tabs at end

### LINKED ISSUE #264: Feature request: open files at the end
Would it be possible to add an option for having files that are opened be added to the end of the tabs rather than right after the currently open tab?

anderoonies:
Resolves #264, not sure how in-demand this feature is though


lee-dohm:
Thanks @anderoonies :grinning:

anderoonies:
:grin:

smaili:
Thanks @anderoonies!

PR #268, opened 2016-02-25
(WIP) Interact with pane (not item) to get and change pending state

kuychaco:

PR #274, opened 2016-02-28
Add support for new pending item API

BinaryMuse:
This adds backward-compatible checks for the new pending item APIs, and falls back to using `Item::terminatePendingState`, etc. if they're not supported in the current Atom environment.

Related to atom/atom#10995, this allows `tabs` to continue to provide visual feedback on item pendingness with both the old and new APIs; will eventually be replaced by #268.

/cc @kuychaco

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -92,7 +92,7 @@ class TabBarView extends HTMLElement
 
   addTabForItem: (item, index) ->
     tabView = new TabView()
-    tabView.initialize(item)
+    tabView.initialize(item, @pane)
     tabView.terminatePendingState() if @isItemMovingBetweenPanes
     @insertTabAtIndex(tabView, index)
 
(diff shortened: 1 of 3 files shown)


BinaryMuse:
Per discussion with @kuychaco, we're going to move specs like this one:

```coffeescript
describe "when switching from a pending tab to a permanent tab", ->
  it "keeps the pending tab open", ->
    editor1 = null
    editor2 = null

    waitsForPromise ->
      atom.workspace.open('sample.txt').then (o) ->
        editor1 = o

    waitsForPromise ->
      atom.workspace.open('sample2.txt', pending: true).then (o) ->
        editor2 = o

    runs ->
      pane.activateItem(editor1)
      expect(pane.getItems().length).toBe 2
      expect(pane.getItems()).toEqual [editor1, editor2]
      expect($(tabBar.tabForItem(editor2)).find('.title')).toHaveClass 'temp'
```

into Atom itself; except for the last line, this test is actually exercising the pending item behavior, which now belongs to `Pane`, and should instead be exercised there. That will also fix the failing specs for this package, which are failing because the tests are designed to run against a newer version of Atom, where this behavior was changed (and keeping the tests in core oversteps this issue).

Simply having specs in this package that test that the UI behaves correctly depending on an item's pending state and its changes is enough.

PR #276, opened 2016-03-03
Only add at end for new tabs

anderoonies:
the implementation i wrote [here](https://github.com/atom/tabs/pull/273) was not quite right—it put a tab to the end _any_ time it was inserted, not only if it was a new item.
the result was that any time a tab was dragged it ended up at the end, clearly not the right behavior.
this fixes that by moving the logic up to where _new_ tabs are added, not where tabs are inserted.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -95,7 +95,10 @@ class TabBarView extends HTMLElement
     tabView = new TabView()
     tabView.initialize(item, @pane)
     tabView.terminatePendingState() if @isItemMovingBetweenPanes
-    @insertTabAtIndex(tabView, index)
+    if atom.config.get('tabs.addNewTabsAtEnd')
+      @appendChild(tabView)
+    else
+      @insertTabAtIndex(tabView, index)
 
   moveItemTabToIndex: (item, index) ->
     if tab = @tabForItem(item)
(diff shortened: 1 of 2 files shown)


lee-dohm:
Thanks @anderoonies!

maxnowack:
it seems that this PR broke the tab sorting if I use `ctrl+tab` or `ctrl+shift+tab`

anderoonies:
:disappointed: this has been breaking more than it's been fixing.
problem is adding the new tab at the end isn't changing its index in the pane. adding the tab to the end should call `@pane.moveItem(item, @pane.getItems().length - 1)` to fix that.
i'll write some specs to make sure indexes are being mirrored between tab bar/panes and push up.

lee-dohm:
@maxnowack and @anderoonies <kbd>Ctrl+Tab</kbd> and <kbd>Ctrl+Shift+Tab</kbd> was changed in the latest release to do most-recently-used tab switching by default. See the blog post for more details:

http://blog.atom.io/2016/04/12/atom-1-7-and-1-8-beta.html

It is still possible that this is still broken but since <kbd>Ctrl+Tab</kbd> was specifically mentioned, I suspect it is the MRU switching change most likely.

anderoonies:
@lee-dohm interesting, i wasn't aware those were being changed. 
the problem exists with `Cmd+Tab+}`, too, though. i misunderstood the API for adding items when originally coding this. a PR that fixes this:
https://github.com/atom/tabs/pull/302

PR #277, opened 2016-03-03
Fix behavior of split panes when `alwaysShowTabBar` is false

### LINKED ISSUE #275: Tab bar never hides in splits with disabled "Always Show tab bar"
After splitting the editor and the tab bar appears it will never disappear despite showing only one tab per split. With the `Always show tab bar` setting disabled I expect that splits with one tab will hide their tab bar. Initially upon split creation they do hide it but once it appears it will stay there.

anderoonies:
fixes #275 
makes a couple changes:
1. update logic in `shouldAllowDrag` so that when a split pane has only one tab the tab bar becomes `hidden`
2. add listener so dragging a tab into a pane view unhides the tab bar
hide-tab-bar

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -17,7 +17,9 @@ class TabBarView extends HTMLElement
   initialize: (@pane) ->
     @subscriptions = new CompositeDisposable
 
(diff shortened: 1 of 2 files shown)


anderoonies:
just ran some tests and behavior is wrong if a tab is dragged over then leaves the pane. taking a look at that

PR #298, opened 2016-04-08
Add 'Close Tabs to Left' right-click menu item

swrobel:
There is a Close Tabs to the Right item, so it seemed logical to me that there should be Close Tabs to the Left. I implemented this because I wanted to close tabs to the left and the option wasn't there.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -22,6 +22,7 @@ class TabBarView extends HTMLElement
       'tabs:close-tab': => @closeTab(@getActiveTab())
       'tabs:close-other-tabs': => @closeOtherTabs(@getActiveTab())
       'tabs:close-tabs-to-right': => @closeTabsToRight(@getActiveTab())
+      'tabs:close-tabs-to-left': => @closeTabsToLeft(@getActiveTab())
       'tabs:close-saved-tabs': => @closeSavedTabs()
       'tabs:close-all-tabs': => @closeAllTabs()
       'tabs:open-in-new-window': => @openInNewWindow()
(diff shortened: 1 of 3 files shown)


ZombieHippie:
Close tabs to the right is generally more useful than close tabs to the right, because new files are opened on the right the tabs on the left tend to be the ones that have been opened the longest and are most important. So, this feature may not be pulled into production.

I personally feel that this certainly adds functionality, but I'm afraid that it would confuse me a little bit from a think about how I'm thinking perspective, because close tabs to the right is common and I can quickly read all the tabs that will be closed from left to right, then if I see another option I will think more.

This is preferential I suppose, but I wouldn't add a lot of context menu items, just for the fact that I don't want to need to think any more than I already do about living my life ;-)

swrobel:
I see it completely the opposite - the tabs to the left were opened long ago and thus are stale, and Closing to the Left is the easiest way for me to clean up the cruft.

50Wliu:
If we are to mimic Chrome here (which parts of Atom do, since it's based on it :grin:), Chrome only has a `Close Tabs to the Right` option.

infogulch:
If we're mimicking Chrome's tab behavior we should add Shift+Click and Ctrl+Click behaviors.

* Shift+Click selects all tabs between the currently active tab and the clicked tab.
* Ctrl+Click selects additional tags individually.
* All selected tabs are closed/moved/reloaded together.

:smile:

50Wliu:
@infogulch Feel free to create a new issue for that!

lee-dohm:
@50Wliu Have you tested this already to verify it works as expected?

50Wliu:
@lee-dohm No, I have not.

lee-dohm:
Ok, I'll see if I can take a look.

swrobel:
Been using it since this PR and it's working great for me, but verify away

lee-dohm:
Verified. Thanks for contributing!

Issue #129, opened 2015-04-08
Tab resize should be delayed


izuzak:
_From @karai17 on April 7, 2015 17:2_

If I have a lot of tabs open and I want to close several of them, it can be a little annoying to have to continue adjusting the mouse when I am trying to close tabs as they resize. Chrome does a good job at this: When you close a tab, the other tabs shift to fill the space but they do not resize to fill space for a second or two, giving you the ability to close several tabs in a row quickly without having to move your mouse.

_Copied from original issue: atom/atom#6287_

izuzak:
_From @mnquintana on April 7, 2015 23:41_

:+1: I really like this idea!

izuzak:
_From @philipgiuliani on April 8, 2015 11:19_

Chrome isn't delaying the tab-resize, they are doing it as soon as you remove the mouse from the tabbar. I also would like to see this nice feature :)

For now, just use `CMD + W`.

**Update**
I just looked over this, and the main problem is, that the tab-resize is handled with CSS (flex), so its currently not easliy possible to implement this feature.

This issue should go to the following repo: https://github.com/atom/tabs

karai17:
Would it be possible to move the reflow styles to only activate on mouse leave?

philipgiuliani:
@karai17 no i dont think this is possible. My idea would be to get the current tab-width with javascript on a mouseover and set it as a style. When the mouse leaves it should be removed again and the tabs will get bigger than. Maybe i could try that out :)

karai17:
That sounds like it could work!

philipgiuliani:
I just tried it out and was not very successful so far :/ The tabbar has `display: flex`, so its not possible to set a fixed-width to the tabs. I did a workaround to set the tab-width and remove display: flex from the tab-bar which worked, but not perfectly. The last tab disappears for some reason but i can look into it.

My current problem is that the active tab has another with than the current tab. Thats in Google Chrome not the case. So currently the delaying has no sense because you get interrupted by the active tab. And as soon as you close it the next tab gets active of course.

karai17:
Any idea on how google manages it in chrome? If atom runs on top of WebKit
or blink, could we not snag chrome's tab system? I know of a project,
clover, that embeds windows explorer inside a chrome framework which is
quite novel and works mostly well.

philipgiuliani:
[...]

PR #300, opened 2016-04-12
Resize tabs when mouse moves away from tab bar

danjordan:
Here's an implementation of the Chrome tabs functionality required in issue #129.

Functionality added...

- Each tab has it's maxWidth set to it's current width when the mouse enters the tabbarview
- Each tab has it's maxWidth set to '' when the mouse leaves the tabbarview

### DIFF
--- a/styles/tabs.less
+++ b/styles/tabs.less
@@ -17,6 +17,7 @@
     -webkit-flex: 1;
     max-width: 175px;
     min-width: 40px;
+    transition: max-width .25s ease-in;
 
     &.active {
       -webkit-flex: 2;
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -46,6 +46,8 @@ class TabBarView extends HTMLElement
       'tabs:split-left': => @splitTab('splitLeft')
       'tabs:split-right': => @splitTab('splitRight')
 
+    @addEventListener "mouseenter", @onMouseEnter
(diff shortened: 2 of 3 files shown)


alexandernst:
Is this going anywhere? 20 days and not even a comment from an atom dev, let alone code review.

mnquintana:
This is looking good! Would you mind adding some kind of transition to the tabs so they don't just pop to their new size?

danjordan:
Not sure if you got a notification when I pushed the code, so just letting you know I have added an animation when the tabs move back to max-height.

rageandqq:
Any updates on this feature?

styles/tabs.less
```diff
@@ -17,6 +17,7 @@
     -webkit-flex: 1;
     max-width: 175px;
     min-width: 40px;
+    transition: max-width .25s ease-in;
```
simurai: Not too critical, but how about `ease-in-out`? Then it doesn't stop too abruptly?

simurai:
Tried it out and UX-wise it works great, especially for even sized tabs. 👍 

tabs

mnquintana:
This was added in #344 – thanks again for all your hard work @danjordan! 👏

mnquintana:
This should land in Atom 1.10

adambuczynski:
@danjordan this is the best feature to land in Atom since the first release :D

PR #302, opened 2016-04-15
Correctly adjust pane index when adding with `addNewTabsAtEnd` config

anderoonies:
Fixes a bug found by @maxnowack [here](https://github.com/atom/tabs/pull/276#issuecomment-210360700).
Moving tabs to the end wasn't adjusting their index in the pane, so iterating over them was out of order. 
Adding a new tab at the end calls the pane's `moveItem` method to change its position in the pane, which is then reflected in the tab bar.
The previous specs for this were only examining the tab bar's items, not the pane's items, so I added a spec which ensures that an added item is the last in the pane.

cc @lee-dohm


lee-dohm:
Thanks @anderoonies!

PR #312, opened 2016-04-24
Get entire icon CSS class from file icon service

acontreras89:
`tree-view` and `tabs` were expecting different outputs from the service. Sorry about that :sweat_smile:

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -190,6 +190,8 @@ class TabView extends HTMLElement
       @itemTitle.classList.remove('icon', "icon-#{@iconName}")
 
     if @iconName = @item.getIconName?() or @path? and @iconName = FileIcons.getService().iconClassForPath(@path)
+      # File icons service will return an entire CSS class
+      @iconName = @iconName.replace('icon-', '')
       @itemTitle.classList.add('icon', "icon-#{@iconName}")
 
   getTabs: ->


Alhadis:
Also, there should still be a way of passing multiple classes. [See here](https://github.com/atom/tree-view/pull/825).

Alhadis:
Never mind, I realised this may instead be a simple oversight instead of by design. I've [submitted a pull request](https://github.com/atom/tabs/pull/319). =)

acontreras89:
@Alhadis I'm aware of the issues and pull requests opened for this service. However, I'm just helping @lee-dohm make this possible, he's the one making the decisions :stuck_out_tongue_winking_eye: 

I can say I kept the `icon-` prefix because I wanted to keep the values in `@iconName` the same. I can see why this is inconvenient, but I didn't want to cause any side effects by implementing this service behavior.

I hope we can decide on the following steps this week, maybe add these improvements too :)

Alhadis:
I guess it didn't happen this week... :(

acontreras89:
Sorry @Alhadis, I'm not having much time to spare in contributing lately. This (all `atom.file-icons` stuff) is my top priority for when I have a chance to do some work, though!

Alhadis:
I still remember what it's like to have a job, so I understand.

I just want to make sure this gets addressed before the next major version of Atom, though. =)

Alhadis:
@acontreras89 Would you be alright with me taking care of the Fuzzy-Finder's consumption of the icons service? =)

acontreras89:
@Alhadis sure, go for it! I wanted to make the change myself but as I told you, I'm not spending much time working for atom lately. Besides, I'm already working in making this change available for the `find-and-replace` package.

It should be pretty straightforward. You can use either the `tabs` or `tree-view` pull request as a guideline for the required changes.
Once you have it ready, ping me (and Lee) so we can take a look and provide some feedback.

Alhadis:
Ace, thanks. =)

Aye, I got the impression you wanted to do it, but I remembered you're probably busy as hell. I also wanted to be sure the icon consumption dealt with less-obvious things like multiple classes and prefixing, etc, so yeah.

Alhadis:
[Done!](https://github.com/atom/fuzzy-finder/pull/211)

Well, that was easier than I expected (not saying anything about adding the specs, mind you). Did I miss anything?

PR #319, opened 2016-05-02
Improve logic for consuming file-icons

Alhadis:
Two problems currently exist with the package's handling of file-icons:

1. **Everything is prefixed with `icon-`**  
This is limiting for authors, who may have CSS classes named differently. It's also inconsistent with the behaviour of the `tree-panes` package, which performs no such prefixing on class-names.<br/><br/>
2. **Only one class name can be set by the service**  
Again, this is very limiting for packages that may need to add more than one class at a time (such as one for the icon, and another for its colour).

This PR alters the existing logic so that
- `icon-` prefixes are limited to "core" panels only (e.g., items with a `getIconName` method defined)
- multiple classes can be supplied by services as either an array or space-delimited list

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -187,7 +187,7 @@ class TabView extends HTMLElement
 
   updateIcon: ->
     if @iconName
-      names = @iconName.split /\s+/g
+      names = unless Array.isArray(@iconName) then @iconName.split(/\s+/g) else @iconName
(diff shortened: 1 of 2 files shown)


lee-dohm:
Could you add some tests to ensure we don't accidentally break these enhancements?

Alhadis:
Damn, sorry. Added.

lee-dohm:
Thanks!

Alhadis:
No, thank YOU. :D

Issue #317, opened 2016-05-02
Filenames display with normalized spaces in tabs


jodekirk:
Opening a file with multiple spaces in the filename displays as one space.
It appears to be normalizing whitespace in the display of the filename.
For example a filename with 2 or more spaces displays in the tab as one space:
"811347 1_por.xml"
displays as 
"811347 1_por.xml"

This screenshot shows two filenames where one has multiple spaces and one that has only one space. They look identical in the tabs title.
![993d12f0-0dea-11e6-887b-5dc2f45b54dd](https://cloud.githubusercontent.com/assets/1869259/14962083/9ed58312-1062-11e6-8911-779abded686a.png)

Atom 1.7.2 Windows 7 64bit

Per damieng potential fix is to change the whitespace: nowrap on the .tab .title class to be whitespace: pre-nowrap

PR #324, opened 2016-05-09
Switch styling to preserve whitespace in file names

zoo1:
Fix for #317.

### DIFF
--- a/styles/tabs.less
+++ b/styles/tabs.less
@@ -29,7 +29,7 @@
 
     .title {
       overflow: hidden;
-      white-space: nowrap;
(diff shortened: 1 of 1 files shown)


simurai:
This seems 👍 

- [x] Sequences of whitespace are preserved
- [x] Ellipsis is still shown when text is overflowing

Ben3eeE:
@simurai did this close https://github.com/atom/tabs/issues/317?

simurai:
@Ben3eeE Ah.. thanks. Yes, I think it can be closed. 👍

Issue #248, opened 2016-01-07
Close all tabs does not work


kafeltz:
In (File -> Close All Tabs) do nothing if a tab is not selected (focused).

To un-focus a tab, just click in some file in tree-view and go right in (File -> Close All  Tabs).

You have to click anywhere in any tab to gain focus before going in Close All Tabs.

iKlsR:
Able to reproduce this in 1.4.

stanfieldr:
I think this is an issue with the way commands are designed. They are just using the active element [here](https://github.com/atom/atom/blob/c5a76d4a7d9b7de9c78a534091a1b775bc5e9c9a/src/atom-environment.coffee#L932) when calling menu commands. And then in the command-registry uses the idea of event delegation. So since the tree view never bubbles the event through the pane, it is never notified.

lee-dohm:
Fixed by #325.

kafeltz:
Super cool

PR #325, opened 2016-05-15
Make close all tabs work without a tab focused

stanfieldr:
See #248 for more details


50Wliu:
It looks like this broke some specs, can you look into that please?

stanfieldr:
I changed it so that if you are active on a pane, it closes all pane tabs. If you are focused on something that isn't a pane, it closes all tabs in all panes.

lee-dohm:
Thanks for this!

PR #327, opened 2016-05-19
remove needless prefixes

Connormiha:
This patch removed needless prefixes for `flex`. Chromium supports clear `flex` since version 29


50Wliu:
👍

PR #330, opened 2016-05-26
Ignore addNewTabsAtEnd config when dragging to a different pane

### LINKED ISSUE #329: Moving tabs between panes with Add New Tabs At End
### Prerequisites

* [X] Can you reproduce the problem in [safe mode](http://flight-manual.atom.io/hacking-atom/sections/debugging/#check-if-the-problem-shows-up-in-safe-mode)?
* [X] Are you running the [latest version of Atom](http://flight-manual.atom.io/hacking-atom/sections/debugging/#update-to-the-latest-version)?
* [X] Did you check the [debugging guide](http://flight-manual.atom.io/hacking-atom/sections/debugging/)?
* [X] Did you check the [FAQs on Discuss](https://discuss.atom.io/c/faq)?
* [X] Are you reporting to the [correct repository](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#atom-and-packages)?
* [X] Did you [perform a cursory search](https://github.com/issues?q=is%3Aissue+user%3Aatom+-repo%3Aatom%2Felectron) to see if your bug or enhancement is already reported?

### Description
If you move a tab using mouse drag and drop from one pane to another it will be moved to the end of the tab bar if you have the "Add New Tabs At End" configuration option enabled.

### Steps to Reproduce

1. Open Atom in safe mode.
2. Ensure you have the "Add New Tabs At End" configuration option set to true. This is found in the configuration options for the Tabs package.
3. Split the Atom window to two different panes.
4. Open a bunch of tabs in each pane.
5. Drag and drop a tab from one pane to the other, the moved tab gets added in the end rather than where it was dropped.

add tabs at end move to other pane

**Expected behavior:** Tab to be moved where it was dropped and not to the end. Technically this is a new tab in that pane but it is also dropped in a specific location which I think should take precedence.

**Actual behavior:** Tab is moved to the end when it is dropped.

**Introduced by:** https://github.com/atom/tabs/pull/273

**Partially fixed by:** https://github.com/atom/tabs/pull/276 This only fixed it for tabs moved within the same pane.

cc @anderoonies

#### atom --version

```
Atom    : 1.7.4
Electron: 0.36.8
Chrome  : 47.0.2526.110
Node    : 5.1.1
```

#### apm --version

```
apm  1.9.2
npm  2.13.3
node 0.10.40
python 2.7.11
git 2.8.1.windows.1
visual studio
```

Windows 10

anderoonies:
fixes #329 

thanks to @Ben3eeE for finding this
spec seems exhaustive to me (and fails without the one-line fix).

### DIFF
--- a/spec/tabs-spec.coffee
+++ b/spec/tabs-spec.coffee
@@ -728,6 +728,28 @@ describe "TabBarView", ->
           expect(pane2.activeItem).toBe item1
           expect(pane2.activate).toHaveBeenCalled()
 
+      describe "when addNewTabsAtEnd is set to true in package settings", ->
+        it "moves the dragged tab to the desired index in the new pane", ->
+          atom.config.set("tabs.addNewTabsAtEnd", true)
+          expect(tabBar.getTabs().map (tab) -> tab.textContent).toEqual ["Item 1", "sample.js", "Item 2"]
+          expect(pane.getItems()).toEqual [item1, editor1, item2]
+          expect(pane.getActiveItem()).toBe item2
+
+          expect(tabBar2.getTabs().map (tab) -> tab.textContent).toEqual ["Item 2"]
+          expect(pane2.getItems()).toEqual [item2b]
+          expect(pane2.activeItem).toBe item2b
+          spyOn(pane2, 'activate')
+
+          [dragStartEvent, dropEvent] = buildDragEvents(tabBar2.tabAtIndex(0), tabBar.tabAtIndex(0))
+          tabBar2.onDragStart(dragStartEvent)
+          tabBar.onDrop(dropEvent)
+
+          expect(tabBar.getTabs().map (tab) -> tab.textContent).toEqual ["Item 1", "Item 2", "sample.js", "Item 2"]
+          expect(pane.getItems()).toEqual [item1, item2b, editor1, item2]
+          expect(pane.getActiveItem()).toBe item2b
+
+          atom.config.set("tabs.addNewTabsAtEnd", false)
+
     describe "when a tab is dragged over a pane item", ->
       it "draws an overlay over the item", ->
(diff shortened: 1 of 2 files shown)


spec/tabs-spec.coffee
```diff
@@ -728,6 +728,28 @@ describe "TabBarView", ->
           expect(pane2.activeItem).toBe item1
           expect(pane2.activate).toHaveBeenCalled()
 
+      describe "when addNewTabsAtEnd is set to true in package settings", ->
```
50Wliu: I don't think `in package settings` is necessary here.  Maybe you could do ```when the `addNewTabsAtEnd` option is true```

spec/tabs-spec.coffee
```diff
@@ -728,6 +728,28 @@ describe "TabBarView", ->
...
+        it "moves the dragged tab to the desired index in the new pane", ->
+          atom.config.set("tabs.addNewTabsAtEnd", true)
```
50Wliu: Maybe make this a `beforeEach` right after the describe?  Are there any other tests that test for `addNewTabsAtEnd`?

anderoonies: agreed, right now `addNewTabsAtEnd` specs are just tacked on wherever needed and it's gotten messy as more have been needed. i could extract them all into one place if you'd like

50Wliu: Your call - there's no one right way to organize the specs.

anderoonies: i'm going to say it's probably a good idea to separate those but outside the scope of this PR. i can open another PR this weekend that separates the tests out

Ben3eeE:
Gave this a round of testing and it fixes the problem for me and I can't notice anything else breaking. :+1:

anderoonies:
🙇🏻 thank you so much!!

PR #334, opened 2016-05-31
Stop override the default `delay` values when adding tooltip

### LINKED ISSUE #333: Stop override the default `delay` values when adding tooltip
Hello! 

I want to make tooltips pop up faster. I have a project that contains files with similar names, but in different directories. When I have opened simultaneously a few of them, I cannot see which is which until I click or wait for a tooltip, but it takes up a second. Atom does not support changing the default value of delay in settings, but it can be overwritten by adding

````coffeescript
atom.tooltips.defaults.delay.show = 100
````

to the `init.coffee` file. Unfortunately, it does not affect the delay of the tabs tooltips, because plugin [adds them overwriting the default `delay` value](https://github.com/atom/tabs/blob/b31ae3d293fb2dccd140e69cf91953ce85e286b5/lib/tab-view.coffee#L155). As long as [these values are identical](https://github.com/atom/atom/blob/f48069c5f848b5ae624b110a1df7800e919eb142/src/tooltip-manager.coffee#L48) setting them is redundant.

Can we remove it?

Greenek:
Fixes #333

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -155,9 +155,6 @@ class TabView extends HTMLElement
       @tooltip = atom.tooltips.add this,
         title: @path
         html: false
-        delay:
-          show: 1000
-          hide: 100
         placement: 'bottom'
 
   destroyTooltip: ->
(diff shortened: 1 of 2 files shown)


Greenek:
Sorry, I messed up with branches a little :) I will re-add this pull request later.

PR #337, opened 2016-05-31
Stop override the default `delay` values when adding tooltip

### LINKED ISSUE #333: Stop override the default `delay` values when adding tooltip

Greenek:
Fixes #333

### DIFF
--- a/lib/tab-view.coffee
+++ b/lib/tab-view.coffee
@@ -155,9 +155,6 @@ class TabView extends HTMLElement
       @tooltip = atom.tooltips.add this,
         title: @path
         html: false
-        delay:
-          show: 1000
-          hide: 100
         placement: 'bottom'
 
   destroyTooltip: ->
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Thanks for contributing 🙇. While I can agree that being able to configure the delay could be nice I am 👎 on merging this. All it accomplishes is adding an undocumented way to configure a value that we don't have tests for or want to support right now. I fear that we will break this without realizing it one day and users who found it through this PR and expect us to support it since we merged it.

/cc: @iolsen @mnquintana

iolsen:
I agree, thanks @Ben3eeE!

PR #349, opened 2016-06-21
Fix auto-reveal scrolling when adding new tab at end

anderoonies:
There was an issue with the call to moveItem focusing a temporary tab when adding a new tab at the end. This caused the tree-view to change scroll between the first click and second click of a double click event, meaning the wrong item could be accidentally clicked.

Fixes https://github.com/atom/tree-view/issues/867

Thanks to @Ben3eeE for finding this

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -100,9 +100,9 @@ class TabBarView extends HTMLElement
     tabView = new TabView()
     tabView.initialize(item, @pane)
     tabView.terminatePendingState() if @isItemMovingBetweenPanes
-    @insertTabAtIndex(tabView, index)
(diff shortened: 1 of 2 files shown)


Ben3eeE:
I have tested this and it does fix the issue for me.
Thanks to @anderoonies for taking the time looking at this!

Also I don't know about the click events firing off the wrong items part as the original issue https://github.com/atom/tree-view/issues/844 stated since I was never able to reproduce it when the tree-view stays in place.

chekalskiy:
Getting this bug for a couple of days

Ben3eeE:
Thanks for contributing 🙇 

As mentioned in https://github.com/atom/tree-view/issues/867#issuecomment-450266632 this is no longer an issue so we are going to close this PR.

PR #374, opened 2016-09-10
Fix invalid reference to item field and remove unused variables

Ingramz:


50Wliu:
> No description provided.

What does this fix?  Are there any open issues pertaining to this? etc.

Ingramz:
It's a mistake or unintended feature that coffeescript allowed in older versions. Newer versions of coffeescript transpile the `item` parameter as `item1` to avoid naming conflict with the line that was changed. If you use an older version of coffeescript, it fixes nothing, in newer versions the code would fail.

Ingramz:
Also removed some unused variables that were found with coffeescope. I verified that the "invalid" reference was the only one.

50Wliu:
Thanks!

Issue #138, opened 2015-05-11
Option for auto-scroll when hidden.


v3ss0n:
Neeed an option for this:
- When tabs are hidden due to too many files  , automatically scroll to the latest , newly opened files.

v3ss0n:
anyone interested in working on this?

CadenP:
Focusing a tab that is already open (selecting a file either by using Fuzzy Finder or Project Find [`CMD+SHIFT+F`]) does not move the tab into view if it is off-screen (too many tabs).

STEPS:
1. Have too many tabs open - let them go off the screen.
2. Use Fuzzy Finder to go to a tab that is off screen.
3. Notice:
 - Tab content is focused
 - Tab itself is focused but still off-screen.

Desired behaviour:
Scroll row of tabs to make focused tab visible.

Actual behaviour:
Doesn't do that. (stays put)

CadenP:
In addition:

`CMD+OPTION+ARROW` to switch tabs continues off-screen.

STEPS:
1. Have too many tabs open - let them go off the screen.
2. `CMD+OPTION+ARROW` repeatedly until focusing on a tab that is not visible.

Desired behaviour:
Tab is focused.
Row of tabs scrolled to make focused tab visible.

Actual behaviour:
Tab is focused.
Tab remains off-screen.


I wonder if some sort of tabOnFocus function should be used to move the whole thing into view... Would solve both of these issues. Would be better than an external function trying to move things around in reaction to to some set of events.

alexmnazarenko:
I second this.

I often work with a lot of opened files (and I need them open, e.g. doing some mass refactoring).

1. Tabs go off-screen, 
2. I don't know what tab is next/prev
3. I get lost and cry

ocoka:
+1

v3ss0n:
@as-cii  can you look into this? since @izuzak  is not active here anymore and , none of the community found a way to fix it yet.

lee-dohm:
This is something we want to fix eventually, not by having an option, but by making it the default behavior. As far as fixing this now, this is of fairly low priority for us relative to the many, many bugs, things that need to be made more performant and other improvements Atom requires. This is something that we would accept a PR for, so long as it was made the default behavior and did not add another option.

Ben3eeE:
This has been bothering me as well so I made an attempt to add this in https://github.com/atom/tabs/pull/375

v3ss0n:
@Ben3eeE  its working very well! thank you very much.
Please merge this

PR #375, opened 2016-09-12
Automatically scroll to tab if not completely visible when activating a pane item

Ben3eeE:
Closes https://github.com/atom/tabs/issues/138

When a tab is out of view automatically scroll to that tab when activated. Current behavior is to scroll so the left or right edge is shown depending on direction.

Gifs of behavior says more than 1000 words.

## New behavior
new activiate out of view
new activiate out of view other dir
ctrl tab

## Old behavior
old activiate out of view
old activiate out of view other dir

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -124,15 +124,6 @@ class TabBarView extends HTMLElement
     tab.updateTitle() for tab in @getTabs()
     @updateTabBarVisibility()
 
-  scrollToTab: (tab) ->
-    tabRightEdge = tab.offsetLeft + tab.clientWidth
-    tabBarRightEdge = this.scrollLeft + this.clientWidth
-
-    if tabRightEdge > tabBarRightEdge
-      this.scrollLeft = tabRightEdge - this.clientWidth
-    else if this.scrollLeft > tab.offsetLeft
-      this.scrollLeft = tab.offsetLeft
-
   updateTabBarVisibility: ->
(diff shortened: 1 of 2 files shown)


v3ss0n:
tested and working very well.
Please merge , that is only thing having a drawback using atom.

Well done! @Ben3eeE

Ben3eeE:
Been using this on and off. There are a few things that reset or scroll the tab bar that gets frustrating when you expect them to be in view.

* Sometimes (Had luck reproducing first save after reload) saving with the `atom-beautify` package installed.
* Saving `styles.less`.
* Reloading the Atom window/Starting Atom. The activate happens too early so it gets measured incorrectly or something.
* Running the `dev-live-reload:reload all` command.

Also this doesn't work the first time an `image-view` is opened I have no idea why that would be. Only the first image opened in that window is incorrect all others seem fine.
image-view

All in all it is an improvement and I much rather work with this than without. But there are many small things that need to be ironed out to make it perfect.

Ben3eeE:
`Element.scrollIntoView` does not scroll if half the tab is visible. So maybe the manual implementation is better...

nathansobo:
@Ben3eeE I definitely think the manual implementation is better if the DOM APIs don't deliver a 100% experience. Tried to push a revert to your branch but I don't have access.

nathansobo:
Thanks @Ben3eeE! This is a really nice touch. Sorry for stealing your PR again @iolsen :trollface:.

PR #380, opened 2016-10-06
Resize tab width on dragLeave

### LINKED ISSUE #371: Tabs do not resize when dragging tabs between panes
### Description

The resize tabs when mouse moves away from tab bar functionality does not resize the tab bar correctly when the tab is moving between panes. The tab bar stays at non full width until you mouse over to trigger the event.
### Steps
1. Open Atom in safe mode.
2. Split right and open a bunch of tabs to the right.
3. Drag tabs from the right to the left.
4. Observe that the tabs to the right never goes back to full size after dropping the tab in the left pane.
### Excepted

Tab bar to resize back to full width when the tab has been moved to another pane because it is away from the tab bar.
### Actual

Tab bar does not resize properly unless you move the mouse over there to trigger the event.
### Gif

tab resize when closed
### Refs

https://github.com/atom/tabs/pull/344
### Versions

Atom 1.10 Windows 7

anderoonies:
Fixes #371 
rezie-dragleave


Ben3eeE:
Thanks for looking at this @anderoonies.

I found another similar issue, can you fix this here in addition to #371 or should I file it in a separate issue?

open in new window

PR #382, opened 2016-10-08
Account for dropping in another pane container

matthewwithanm:
In Nuclide, we're using additional PaneContainers to give people more control over their workspace.  Everything works fine for the most part, but you can't drag into the tab bar because the `fromPane` is assumed to be in the same PaneContainer. This commit removes that assumption and falls back to searching the document for the pane.

tabs2


iolsen:
👍 on the new functionality here. If you'll add a test or two covering the new block we'll merge this.

iolsen:
👌 looks good, thanks!

matthewwithanm:
❤️ Thanks for the quick merge!

PR #388, opened 2016-10-17
List view and option for MRU tab switching

iolsen:
Adds a list view of the MRU tab stack to make traversing it easier and more predictable.

mru-basic-list

Part of the fix for https://github.com/atom/atom/issues/11035.
Requires https://github.com/atom/atom/pull/13008 and https://github.com/atom/tabs/pull/388.
Fixes https://github.com/atom/atom/issues/11439


Ben3eeE:
Applies to some of https://github.com/atom/atom/issues/11439.

How does the list look when there are multiple files with the same name?

iolsen:
> How does the list look when there are multiple files with the same name?

Yeah this isn't handled well yet. I think the finished version looks just like the fuzzy finder list, with a second row to disambiguate duplicate names, and file/status icons.

thomasjo:
Might be wise to (visually) limit the size of the list? Just curious what it would look like if someone has 30+ files open. Although I guess this might already be handled via `atom-ui` which this is possibly leveraging automagically?

iolsen:
>Might be wise to (visually) limit the size of the list? Just curious what it would look like if someone has 30+ files open.

Yeah I have something like this in mind. Probably just a hard limit of ~20.

iolsen:
Aside from figuring out why tests are failing on CI but not for me locally, this is just about ready:

mru-ux-2

I tried to make this look like other file selection modals without reusing SelectListView as that's in the process of being removed:

- It has the second line to disambiguate duplicate file names in different directories.
- It uses the same set of default icons as fuzzy-finder and consumes the file icon service
- It has the same git status icons as fuzzy-finder et al
- Rather than choosing an upper limit I simply have the list scroll, just like fuzzy finder.

One minor thing that's not working is that the scrollbar doesn't appear when scrolling the list. I'm pretty sure this is a CSS thing (as opposed to a jquery thing). @simurai do you see something I've missed that's preventing the scrollbar from being visible while scrolling?

Ben3eeE:
Been using this the entire workday and loving it, two small things I noticed:

* Untitled buffers write undefined as the path
image
* If the window loses focus during a partial match and you release the <kbd>Ctrl</kbd> key while it doesn't have focus the key release never gets registered so the List View is stuck open waiting for a <kbd>Ctrl</kbd> key release event. Simply tapping the <kbd>Ctrl</kbd> gets Atom unstuck again.

iolsen:
Nice feedback, thanks. I'll fix the undefined titles, obviously.

As for the behavior not catching the ctrl keyup when Atom doesn't have focus, that's working as intended. Fixing this would require (another) pretty complex change to the keymap. I think this is a minor quirk that's still highly usable and not worth trying to "fix."

simurai:
@iolsen Scrollbar should be fixed.

It's super strange, the reason it works for the fuzzy finder is because there is the `<atom-text-editor>` above. `¯\_(ツ)_/¯`

Ben3eeE:
Continuing the list of things I notice:

* The `settings-view` uses the wrong icon. It should use the icon defined by the `settings-view` instead of the default no icon icon. Same goes for `Project Find Results` and `Markdown Preview` and other packages that define an icon. See images and setting in the `tabs` package to use this icon for the tab bar:
image
image
* I think not making complex changes to the keymap right now is a great idea :laughing:. But perhaps we should allow the MRU list to be canceled by clicking outside it with the mouse or with the <kbd>Esc</kbd> key. Discovering that you have to tap <kbd>Ctrl</kbd> (Or whatever your binding is) might not be very intuitive the first time.

iolsen:
Thanks for all the great feedback. I'll implement both of those.

iolsen:
This requires https://github.com/atom/atom/pull/13549. Once that's merged this is ready to go to finally fix https://github.com/atom/atom/issues/11035.

Note that the newly added MRU setting defaults on but has no affect for any users that have overridden ctrl-tab or ctrl-shift-tab via a manual user keymap tweak or a package installation. It won't annoy anybody who's already turned the feature off the hard way.

iolsen:
screen shot 2017-01-06 at 11 43 07 am

amrocha:
Hey, I love MRU tabbing but I can't stand the tab list. I looked for a way to disable just the list but couldn't find it, if I made a PR to add an option to disable just the list, would you consider merging it?

iolsen:
> if I made a PR to add an option to disable just the list, would you consider merging it?

Yes, we would!

amrocha:
Alright, I'm working on something and will open a PR here when I'm done

timtjtim:
@amrocha Did you manage to do that PR?

Ben3eeE:
@timtjtim There is a pull request open, unfortunately it got abandoned before it was finished /cc: https://github.com/atom/tabs/pull/422

amrocha:
@timtjtim Ben's right, this totally fell off my radar and I've just been using my branch locally, sorry about that :(

I'll try to get around to it but if you have time to help out with rebasing/tests I'd really appreciate that

timtjtim:
I’m afraid I would be no help with that - I’m not much of a programmer. It would be great if this feature was added though - but it’s no rush!

PR #397, opened 2016-11-10
Refresh tab-icons when changing item's title

Alhadis:
A blank editor doesn't update its tab's icon after it's been saved to disk. Meaning that the file-icon service doesn't have a chance to show anything in the tab until it's been closed and reopened.


spec/tabs-spec.coffee
```diff
@@ -1351,3 +1353,72 @@ describe "TabBarView", ->
 
         runs ->
           expect($(tabBar).find('.tab:eq(1) .title')).toHaveClass "status-modified"
+
+describe "File-icons service", ->
```
50Wliu: In other packages this has been added as a standalone spec file `file-icons-spec.coffee`.

50Wliu: He knows better than I do :)

Alhadis: Hence why I'm relying on precedent scenarios. =)

I'm hoping #966 gets reviewed sooner, though. The hacks I'm writing to monkey-patch the tree-view's file-moving routine is nauseating.

Alhadis:
Fixed in #402.

PR #399, opened 2016-11-17
Added option to quickly copy file path

stanfieldr:


stanfieldr:
turns out there's a package that does exactly this

PR #402, opened 2016-12-15
Remove atom-space-pen-views and custom elements registration

as-cii:
This pull request contains an architectural change that replaces the usage of atom-space-pen-views and custom elements respectively with raw DOM APIs and a JS object managing a non-custom DOM element internally.

As always, the scope of the change was non-trivial, so there's a risk we might have introduced some regressions. @ungb: it would be awesome if you can help preventing this by extensively testing this package (and check if popular packages that interact with `tabs` throw exceptions that could be related, e.g. calling a method that does not exist, calling a method or getting a property from a null object, etc.). ✨

### DIFF
--- a/lib/html-helpers.coffee
+++ b/lib/html-helpers.coffee
@@ -1,15 +1,4 @@
-closest = (element, selector) ->
-  return unless element?
-  return element if element.matches(selector)
-  closest(element.parentElement, selector)
-
-indexOf = (element, elements) ->
-  elements ?= element.parentElement.children
-  for child, index in elements
-    return index if element is child
-  return -1
-
 matches = (element, selector) ->
   element.matches(selector) or element.matches(selector + " *")
 
-module.exports = {matches, closest, indexOf}
+module.exports = {matches}
--- a/lib/html-helpers.coffee
+++ b/lib/html-helpers.coffee
@@ -1,4 +1,0 @@
-matches = (element, selector) ->
-  element.matches(selector) or element.matches(selector + " *")
-
-module.exports = {matches}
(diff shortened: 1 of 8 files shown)


ungb:
Hey @as-cii I'm still testing this, but I noticed that everything is a lot slower when I link the tabs package with your changes.  When I unlink it Atom is a lot faster.  Here's a screenshot of before and after:

beforetab
aftertab

as-cii:
I think it might be due to linking the package, which forces Atom to transpile its assets (e.g. Coffeescript, Javascript, Less files, etc.). After loading Atom a second time, the slowdown should disappear.

Based on our Slack conversation, I'll go ahead and 🚢 this because no regression has been found. Thanks so much for you help on this, @ungb! 🙇

Alhadis:
@as-cii This broke the ability to manually refresh tab-icons using `updateIcon`. See #397.

ungb:
@alhadis,  thanks for keeping and eye out. . Hopefully it isn't an issue.    Let us know if you do find any other issues. I tested mainly general use cases and other packages that interact with tabs, but might of missed somethings. Great to always get additional eyes and feedback :)

Alhadis:
I did notice it's not adding an icon class of `icon` when updating. The consumed icon-class is being applied to the `.itemTitle` property, but not the actual `icon` class itself.

as-cii:
> I did notice it's not adding an icon class of icon when updating.

Nice catch, @Alhadis! ✨ Can you point out the exact DOM element that we should apply the `icon` class to?

Alhadis:
Whatever element receives the icon classes...

as-cii:
> Whatever element receives the icon classes...

Interesting, looking at the diff it doesn't seem like I changed anything related to that: before this pull request the `icon` class was being applied only to `@itemTitle`. What am I missing?

Alhadis:
Doesn't matter. We're not using your icon service anymore anyway.

PR #419, opened 2017-03-10
Decorate docks with tabs

matthewwithanm:
This updates tab decoration to also apply to docks (atom/atom#13977)

**Not ready to land yet**


matthewwithanm:
Actually, this won't be necessary anymore since `workspace.observePanes()` will return all the panes. Closing!

PR #420, opened 2017-03-16
Fix tests that docks feature broke

matthewwithanm:
atom/atom#13977 adds more pane containers to the workspace, causing some of this package's tests to fail. This commit updates the tests to be a little more specific.

Not that this it doesn't add new tests to exercise dock tabs; that'll come in another PR.


maxbrunsfeld:
Thanks!

PR #422, opened 2017-03-23
Added a setting to enable/disable the MRU tab list

amrocha:
### Description of the Change

Added a setting to enable/disable the tab list that appears when using MRU tab switching. In this implementation the list is coupled to MRU tab switching. I didn't make it a separate setting since the point of the list seemed to be to give more context when using MRU, and it would be annoying to have a list when using sequential tab switching.

The code design is fairly simple and just plugs in to what was already in place for MRU tab switching.

### Alternate Designs

I thought about making the tab list an independent setting from MRU but decided against it after reading the discussion around the feature.

### Benefits

You'll be able to decide whether you want the tab list or not! User choice!!

### Possible Drawbacks

I don't think so but too much choice???

### Applicable Issues

Didn't see any

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -12,7 +12,8 @@ module.exports =
     @mruListViews = []
 
     keyBindSource = 'tabs package'
-    configKey = 'tabs.enableMruTabSwitching'
+    mruConfigKey = 'tabs.enableMruTabSwitching'
+    listConfigKey = 'tabs.enableMruTabList'
 
     @updateTraversalKeybinds = ->
       # We don't modify keybindings based on our setting if the user has already tweaked them.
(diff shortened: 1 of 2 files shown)


amrocha:
I still need to write tests for this, will do that later today

raffomania:
For anyone interested in a quick fix, I've gone and worked around this using `styles.less`:
```
.modal.overlay.from-top {
    padding: 0;
}

.tabs-mru-switcher {
    display: none;
}
```

iolsen:
This functionality was implemented by https://github.com/atom/tabs/pull/388.

iolsen:
I misunderstood. This is to keep the functionality but disable the visible list.

iolsen:
Though this should probably get closed if it's abandoned. :)

@amrocha are you still hoping to add the tests necessary to get this merged?

iolsen:
I was in the neighborhood this afternoon so I added this setting (in a simpler way, I think) to #529.

PR #424, opened 2017-03-28
Allowed Locations & Permanent Dock Items

matthewwithanm:
These address the feedback in atom/atom#13977 by:

1. Only adding the close icon for dock item when `isPermanentDockItem()` returns `false` (or is absent).
2. Only allowing an item to be dropped into tab bars and panes if its `getAllowedLocations()` method includes the location (or is absent).

I updated layout to not show the overlay over disallowed locations. It would be nice to also change the cursor from copy to `no-drop` then too, but I couldn't find where this was being done.

### DIFF
--- a/lib/layout.coffee
+++ b/lib/layout.coffee
@@ -13,7 +13,8 @@ module.exports =
     @lastCoords = e
     pane = @getPaneAt e
     itemView = @getItemViewAt e
-    if pane? and itemView?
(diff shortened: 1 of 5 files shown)


matthewwithanm:
(tabs are still closable via the context menu and commands)

maxbrunsfeld:
👌

PR #431, opened 2017-04-05
Don't show placeholder when dragging an item over a disallowed location

maxbrunsfeld:
/cc @matthewwithanm


matthewwithanm:
❤️

Sorry, should've been more thorough with e0237e207ab3266ee920f257b7a7877d7ea4b0a8 😩

PR #437, opened 2017-05-05
Fix mistakenly shown docks drop indicator on Electron >= 1.4

as-cii:
This pull request fixes a regression that we were observing on newer Electron versions due to reasons similar to the ones described in #426. In particular, on Electron >= 1.4 we were showing a drop indicator even for dock locations that were unsupported by the dragged item.

Analogously to the aforementioned pull request, in this one we are replacing usages of `dataTransfer.getData` in the `dragover` handler with a scan of the keys present inside the `dataTransfer` object.

/cc: @nathansobo @maxbrunsfeld


nathansobo:
Thanks!

PR #439, opened 2017-05-12
Delay activation of item on mousedown event

matthewwithanm:
Because the item's activation could result in a change of focus, we
need to wait to take it until the event has finished propagating.
Otherwise, the focus will be set by the browser as part of the default
action for the mousedown event.

Sorry there's no test but it doesn't seem possible to simulate the problematic browser behavior and still test what we want to.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -404,8 +404,14 @@ class TabBarView
       @rightClickedTab.element.classList.add('right-clicked')
       event.preventDefault()
     else if event.which is 1 and not event.target.classList.contains('close-icon')
-      @pane.activateItem(tab.item)
-      setImmediate => @pane.activate() unless @pane.isDestroyed()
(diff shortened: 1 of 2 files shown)


matthewwithanm:
The failure must be related but it doesn't repro for me on master or 1.17-beta5 😖

matthewwithanm:
Okay so @zertosh just told me that it actually does make sense that the timeout could fire before `setImmediate()` (TIL). Using the same scheduler makes us green. (The promise brings in the microtask queue too, but that's not important…I just did it because we don't have a `waitsForCallback` 😝)

as-cii:
This looks good, @matthewwithanm! Thanks for this pull request. ✨ 

I only tweaked tests a little bit in a295987 to rely more on the state of the system rather than waiting for the next `setImmediate` callback. I will 🚢 this as soon as we get a green build.

PR #440, opened 2017-05-12
Tab bar shouldn't be focusable

matthewwithanm:
This was originally added way back in 310c058 however, a tabindex of -1
means that the tab bar can receive focus.

As a result, focus would end up being transferred to the tab bar as
part of the default action of the mousedown handler. When the `Pane` is
activated, it would see that its child (the tab bar) had focus and
leave it there, meaning the tab bar (instead of the editor) would
retain focus. This behavior was first reported in #14173, however the
original fix for it (#14175) resulted in Panes always stealing the
focus from their children. This regression was rectified with
atom/atom#14403 and atom/tabs#439, but really the tab bar shouldn't be
focusable in the first place. Making this adjustment means we no longer
have to delay the activation of the item and pane (atom/tabs#150 and
atom/tabs#tabs#439) since the focus will no longer be given to the tab
bar as part of the default action of the mousedown handler.

This shouldn't be merged until after 1.17 leaves beta. For the 1.17 beta, atom/atom#14403 and atom/tabs#439 are the less invasive fixes.

cc @ungb

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -13,7 +13,6 @@ class TabBarView
     @element.classList.add("tab-bar")
     @element.classList.add("inset-panel")
     @element.setAttribute('is', 'atom-tabs')
-    @element.setAttribute("tabindex", -1)
 
     @tabs = []
     @tabsByElement = new WeakMap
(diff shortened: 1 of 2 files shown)


lee-dohm:
Resolve conflicts and ensure the builds are green then this can be merged! :+1:

iolsen:
Ping. If you can update this so it merges, we'll merge it.

matthewwithanm:
hey! it looks like almost all of this work was done done by the PR that switched the interaction from mousedown to click, except removing the tabindex attribute. there's probably no harm in removing the tabindex but, given the other change, there isn't really a motivator for doing it either so i'm going to close.

PR #454, opened 2017-06-13
Fix handling of dragging pane items over tab bars

maxbrunsfeld:
In order to add a tab bar to every pane, we iterate over every pane container on startup and call `observePanes`. Previously, every callback passed to `PaneContainer.observePanes` captured the same `location` variable from its containing scope. This caused every `TabBarView` created after startup to have its `location` set to `bottom` (the final value of the `location` variable).

This bug became more apparent after we disallowed putting TextEditors in Docks, because it meant that you couldn't drag editors between panes: TextEditors are only allowed in the `center` location, but every `TabBarView` thought it was in the `bottom` location.

I've adjusted the aforementioned code to iterate over the pane containers using `Array.forEach` so that we avoid sharing variables between the pane containers' `observePanes` callback.

Also, while I was here, I got rid of some usages of the old `atom.views.getView` API.

Refs https://github.com/atom/tabs/pull/424

/cc @matthewwithanm

### DIFF
--- a/lib/layout.coffee
+++ b/lib/layout.coffee
@@ -1,7 +1,7 @@
 module.exports =
   activate: ->
     @view = document.createElement 'div'
-    atom.views.getView(atom.workspace).appendChild @view
+    atom.workspace.getElement().appendChild @view
     @view.classList.add 'tabs-layout-overlay'
 
   deactivate: ->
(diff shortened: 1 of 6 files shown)


matthewwithanm:
👍 blarg, good catch!

PR #467, opened 2017-06-26
[RFC] Add layout to fit mode ala Sublime

### LINKED ISSUE #136: Tab Bar: Tabs are hidden if too many tabs opened.
_From @v3ss0n on May 11, 2015 8:17_

atom 198 behanvior make tabs more readable but when too many tabs are opened , new ones are hidden. Previously atleast an icon width is still shown.

I propose a solution to add a Next/Back icon at the side of tab strip if too many tabs are opened like in firefox , with additional tablist button.

If i am good with coffeescript i can help.

_Copied from original issue: atom/atom#6730_

xixixao:
### Requirements

Tests TBD

There are 2 Todos, feedback would be appreciated.

### Description of the Change

Support stacking of tabs in the tab bar ala Sublime Text. When in this new mode, we listen on mousewheel, keeping track of the virtual scroll position. We update each tabs absolute position. We make sure that the layouting is redone on any resize (window resize, dock resizing by drag, pane resizing by drag, opening child panes, closing child panes; opening/closing of docks cannot be supported atm. as there is no event for them). Also preserves the "show active tab on selection behavior" (same way as in Sublime). Also preserves drag and drop behavior (anywhere, including at the stacked ends).

screen shot 2017-06-26 at 23 30 32

### Alternate Designs

Maybe as a separate package? Not sure if that is doable though.

This doesn't implement Sublime's gradual decreasing of the shown offset of each tab, the advantage is that if your active item's tab is far on either side, it will still be visible. The downside is that the sides take up more space (at least that still incentivizes you to close your tabs eventually).

### Benefits

The current implementation of the tab bar is pretty useless if you have long file names. Additionally if you have many tabs open (even with short names), and you scroll the tab bar away, you lose the active item's tab (which you might hold as a "mental position" from which you find other tabs). With this behavior, you can always find the tab for the active item, plus you always see how many tabs you have open in total, and you always see the tabs' full names.

### Possible Drawbacks

There is one more option now (the one that enables/disables this mode). Perf-wise this change should be negligible (and enabled runs smoothly at least on my machine).

### Applicable Issues

Fixes #136

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -69,19 +69,27 @@ class TabBarView
 
     @subscriptions.add @pane.onDidAddItem ({item, index}) =>
       @addTabForItem(item, index)
+      @recalculateLayout()
 
     @subscriptions.add @pane.onDidMoveItem ({item, newIndex}) =>
       @moveItemTabToIndex(item, newIndex)
+      @recalculateLayout()
 
     @subscriptions.add @pane.onDidRemoveItem ({item}) =>
       @removeTabForItem(item)
+      @recalculateLayout()
 
     @subscriptions.add @pane.onDidChangeActiveItem (item) =>
       @updateActiveTab()
 
+    @subscriptions.add @pane.onDidChangeFlexScale =>
+      window.requestAnimationFrame =>
+        @recalculateLayout()
+
     @subscriptions.add atom.config.observe 'tabs.tabScrolling', @updateTabScrolling.bind(this)
     @subscriptions.add atom.config.observe 'tabs.tabScrollingThreshold', => @updateTabScrollingThreshold()
     @subscriptions.add atom.config.observe 'tabs.alwaysShowTabBar', => @updateTabBarVisibility()
+    @subscriptions.add atom.config.observe 'tabs.layoutToFit', @updateTabBarLayouting
 
     @updateActiveTab()
 
@@ -166,7 +174,10 @@ class TabBarView
       @activeTab?.element.classList.remove('active')
       @activeTab = tabView
       @activeTab.element.classList.add('active')
-      @activeTab.element.scrollIntoView(false)
+      if @layoutToFit
+        @recalculateLayoutToShow @activeTab.element
+      else
+        @activeTab.element.scrollIntoView(false)
 
   getActiveTab: ->
     @tabForItem(@pane.getActiveItem())
@@ -321,6 +332,8 @@ class TabBarView
         else
           tab.element.parentElement.appendChild(placeholder)
 
+    @recalculateLayout()
+
   onDropOnOtherWindow: (fromPaneId, fromItemIndex) ->
     if @pane.id is fromPaneId
       if itemToRemove = @pane.getItems()[fromItemIndex]
@@ -385,6 +398,10 @@ class TabBarView
   onMouseWheel: (event) ->
     return if event.shiftKey
 
+    if @layoutToFit
+      @recalculateLayout event.wheelDeltaX
+      return
+
     @wheelDelta ?= 0
     @wheelDelta += event.wheelDeltaY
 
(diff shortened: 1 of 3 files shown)


simurai:
This looks like a great addition. 👍  I'll test it out for a bit.

xixixao:
Hey @simurai, any feedback?

simurai:
@xixixao Overall it works great and haven't noticed any big issues. Below just some questions and observations:

1. Is there a specific reason the close buttons (`x`) are hidden? 
2. To position the tabs, `left` is used. Would `transform` perform better? Then no repainting of the tabs is needed. Although I can't see a big difference, probably also depends on the machine and theme. I think one downside with transforms could be that sub-pixel aliasing of the text is lost.
3. Is it a problem that it might break how themes implemented their tabs? Like the One themes have 3 tab modes that have no effect once the "layout to fit" mode is enabled. We could mention "only works if", but that could also confuse and complicate things. And the Atom themes have those skewed sides (like Chrome) that cover the text:
  screen shot 2017-07-11 at 4 41 20 pm
  We should be able to fix it in the Atom themes, but there might be more community themes that have custom styling.

Actually, `#3` is more a question for @atom/core to consider.

xixixao:
1. My bad, I did this because they were showing up more tabs that were hidden. Need to fix this.
2. I haven't seen any performance issues, but I'm not an expert in DOM rendering.
3. Templates will have to adjust. I don't think it breaks them, as it's an optional flag, but you can see in my TODOs that a part of this should be handled by the theme. This kind of extends the "contract". Is the contract only part of `core`?

simurai:
> 3. Templates will have to adjust. I don't think it breaks them, as it's an optional flag

Right. It's just a problem if somebody enables this option and then at a later point installs a theme or package and it will be broken because no support for "layout to fit" has been added. Then the user thinks something is wrong with the theme/package and probably wouldn't try to disable "layout to fit".

Again, this isn't just a problem with this PR and we have to think what to do in other cases too. I'll try to bring it up with the team.

xixixao:
@simurai Any progress? Can we launch this as an alpha, adding messaging that this might not work with all themes?

simurai:
I talked to the team and it's ok to break themes and packages, and at some point can't be avoided, but we would have to see if cost/reward is justified. Not always easy to decide.

> Can we launch this as an alpha, adding messaging that this might not work with all themes?

We used "Experimental" for the custom title bar setting. Somewhat a "Some features will be compromised and use at your own risk".

Anyways, have you considered launching this feature as a separate package? Either as add-on. Or if that's not possible, as a replacement. People would disable the bundled tabs package and install `tabs-plus` or similar where "Layout to fit" is already the default. Then we could see how people like it and if there are bigger issues.

xixixao:
I don't know how to implement this as an addon. This pull request is fairly small but embedded in the current implementation, do you know how to make it an addon (any examples)?

I could launch a replacement, but that would be a fork of this project. All improvements/fixes that go here I'd have to track or they'd be lost. I also think this would pose high barrier to adoption, so you wouldn't really get the information on whether people like it.

Would it be possible to merge, and if people don't find it useful, remove it again?

lee-dohm:
> Would it be possible to merge, and if people don't find it useful, remove it again?

This is generally problematic because once something is in the product people get very upset if it is later taken away. I'm also not a fan of the "experimental" label option because people don't read it.

>  do you know how to make it an addon (any examples)?

The [multirow-tabs package](https://atom.io/packages/multirow-tabs) might be helpful for that.

xixixao:
I will look into implementing this as an addon (possibly with some smaller changes to this package if I need to expose any internals), and fixing the remaining issues.

xixixao:
This is now on `apm` as [stacked-tabs](https://atom.io/packages/stacked-tabs).

It supports both One and Atom themes and [pinned-tabs](https://atom.io/packages/pinned-tabs).

I had to use some monkey-patching and going into the internals here to make it work. Hopefully it can be merged back in if people find it useful.

nathansobo:
@xixixao if you want to add a service interface formalizing and API around what you monkey patched I'd give it a look. If you're interested, let's sketch about what it would look like before you have to spend time coding. Thanks for your understanding on this. Trying to strike a balance between moving ahead and avoiding theme breakage.

PR #512, opened 2017-08-13
Provide service API

UziTech:
### Description of the Change

Provided a service that lets other packages get the path of a context target in a standard way.

At the moment any package that interacts with the tabs package to get files or other information are doing so by getting the active package's `mainModule` or checking classes and data attributes on elements.

There is a standard way to get this information through the [Services API](http://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/)

### Alternate Designs

We may want to change what actually gets provided.

### Benefits

Other packages can interact with this package in a standard way.

### Possible Drawbacks

none

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -87,7 +87,8 @@ module.exports =
       @updateFileIcons()
 
   provideTabs: ->
-    @tabBarViews
+    isTab: (target) -> target.closest(".tab-bar > .tab")?
+    getPath: (target) -> target.closest(".tab-bar > .tab")?.querySelector(".title")?.dataset?.path
 
   updateFileIcons: ->
     for tabBarView in @tabBarViews
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -86,6 +86,9 @@ module.exports =
       FileIcons.resetService()
       @updateFileIcons()
 
+  provideTabs: ->
+    @tabBarViews
+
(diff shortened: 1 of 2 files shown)


UziTech:
The only thing I can think of using this service for is getting the path of a tab that was right clicked on.

A package like [context-git](https://github.com/UziTech/context-git) just uses the [`data-path` attribute](https://github.com/UziTech/context-git/blob/e40e27acbbd526c88847b0e0cc28402113ff299c/lib/helper.js#L52) from the `.title` element. It seems like there should be a standard way to get the path given a target.

lib/main.coffee
```diff
@@ -86,6 +86,10 @@ module.exports =
       FileIcons.resetService()
       @updateFileIcons()
 
+  provideTabs: ->
+    isTab: (target) -> target.closest(".tab-bar > .tab")?
```
UziTech: We may just want to cycle through `@tabBarViews` and check if `target` is descendant of one of them.

Or some other way?

lib/main.coffee
```diff
@@ -86,6 +86,10 @@ module.exports =
...
+    getPath: (target) -> target.closest(".tab-bar > .tab")?.querySelector(".title")?.dataset?.path
```
UziTech: Again we should probably find a better way to do this. Any suggestions?

PR #515, opened 2017-08-18
Perform left/middle click actions on `click` instead of on `mousedown`

### LINKED ISSUE #198: Tabs should close on middle click, not middle mousedown, like other tabbed applications.
In other applications that use tabs (such as Google Chrome, Firefox, etc), middle clicking tabs closes them. Atom sort of has this feature, but instead of closing the tab on mouseup (which triggers the click event), Atom closes the tab immediately upon mousedown. I was a bit surprised by this behavior, and more than once I've accidentally closed the wrong tab. In other applications where middle clicking a tab closes it, you can change your mind in the middle by dragging the cursor away from the tab while still holding down the middle button. This is the same behavior buttons have with respect to left clicking.

I think that tabs should be closed on middle click, rather than on middle mousedown.

I'm not sure what would go into changing this behavior, so I look forward to hearing feedback on this feature request.

as-cii:
Fixes https://github.com/atom/atom/issues/15197
Fixes #198 
Supersedes and closes #385

Other than simplifying the code (we don't need to schedule a `setImmediate` callback to focus the item corresponding to the clicked tab), this pull-request also serves as a workaround to a Chromium bug that blocks the UI thread when middle-clicking a tab on Linux and showing a message box in the handler.

Using `click` slightly changes the UX, because left and middle click actions will be performed when users actually end the click action, rather than when they start it. Chrome, Safari and the macOS UI all behave this way, so I believe this is a good tradeoff from a user-experience standpoint.

Please, note that we still need to handle the right-click event via a `mousedown` handler, because Chromium does not seem to fire `click`events when opening context menus.

@matthewwithanm: I saw you introduced the `setImmediate` scheduling for focus reasons. With these changes, pane activation seem to exhibit the correct behavior, so I went ahead and deleted that workaround. However, I'd ❤️ to get your feedback on this in case you think it's not quite we want.

@Ben3eeE @ungb @50Wliu: as mentioned above, this slightly changes the UX and rearranges the way we respond to mouse input events in tabs. Can you help me test this doesn't regress any unrelated feature? In particular, it may be good to test all the mouse interactions on the various platforms and ensure everything works as expected. I performed some tests on Ubuntu and macOS, but some more QA on this would be super appreciated. ✨ 

/cc: @nathansobo for 👀

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -86,6 +86,7 @@ class TabBarView
     @updateActiveTab()
 
     @element.addEventListener "mousedown", @onMouseDown.bind(this)
+    @element.addEventListener "click", @onClick.bind(this)
     @element.addEventListener "dblclick", @onDoubleClick.bind(this)
 
     @onDropOnOtherWindow = @onDropOnOtherWindow.bind(this)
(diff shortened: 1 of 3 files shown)


matthewwithanm:
@as-cii Yeah, the `setImmediate()` scheduling was already there, but only for the pane activation…I just moved the item activation into it and added the comment (#439).

All of that was specifically to work around the mousedown behavior, which I figured we had to keep since it was intentionally added in #124 but I don't really have any opinions about it. I also opened #440 which got rid of the need for the timeout but it kinda fell through the cracks.

Ben3eeE:
LGTM ⚡️

I am unsure about changing the switch tabs on mousedown behavior in a hotfix release.

nathansobo:
I think we should take the risk considering how terrible this is for Linux users.

robwierzbowski:
Did this revert the performance improvements of https://github.com/atom/atom/issues/5174 / https://github.com/atom/tabs/pull/124? As a new Atom user coming from Sublime, waiting for `mouseup` (complete click) for tab switch feels very sluggish. 

Agree, should click for some of the other actions I see in here (tab close?), but I feel for tab switching this is a regression.

PR #523, opened 2017-09-24
Resize tab width onDragLeave or when tab is opened in a new window

### LINKED ISSUE #371: Tabs do not resize when dragging tabs between panes

50Wliu:
### Requirements

### Description of the Change

Continuation of #380, with @Ben3eeE's comments addressed and the code updated.

### Alternate Designs

I considered adding a global variable `mouseActive` to track if the mouse was in the tab bar or not, but it looks like `mouseleave` is buggy and does not fire a) when dragging or b) when right-clicking.

### Benefits

"Smoother" animations.

### Possible Drawbacks

This same exact code snippet now appears three times in the codebase.

### Applicable Issues

Fixes #371
Supersedes and closes #380

@Ben3eeE for testing


Ben3eeE:
LGTM! :shipit:?

Issue #361, opened 2016-08-04
Add delay to MRU when switching tabs


wprater:
I often switch from a tab that I can see is a few away very quickly and then I want to go back to my original tab I was editing on.  I generally use the fuzzy finder to swap through files in MRU order.  but when switching tabs, they are logged into MRU immediately. can we add a small 150ms or so delay before committing them to MRU.  See how ST3 operates for a comparison.

caleb531:
Application switching in MacOS behaves similarly—there is a short (150ms-200ms?) delay between the time the key binding is pressed and the time the app switcher panel shows. I believe this is primarily to avoid the flicker effect of the panel showing and hiding quickly if the user switches tabs quickly via `cmd-tab` or whatnot. Unfortunately, the Atom MRU tab switcher suffers from this annoying flicker effect; adding a delay of 150ms or so should eliminate the flicker (in addition to delaying the MRU logging, as @wprater mentioned).

iolsen:
Fixed in #529.

PR #529, opened 2017-10-17
Improve MRU tab list display

iolsen:
Two MRU display related improvements:

- a 150ms delay before showing the MRU tab list as suggested in #361 
- an option to disable display of the list altogether as previously attempted in #422.

mru-list-display-setting


iolsen:
Fixes #361.

PR #539, opened 2018-05-13
Show tab bar if the setting is true or there is more than one tab

### LINKED ISSUE #275: Tab bar never hides in splits with disabled "Always Show tab bar"

50Wliu:
### Requirements

### Description of the Change

This PR changes `updateTabBarVisibility` to hide the tab bar if there is less than two tabs, regardless of how many panes are open (as the tab bar is per-pane).

Example:
tab-bar-visibility

I inlined `shouldAllowDrag` as it was only being used by `updateTabBarVisibility`, and not even for dragging purposes.

~~This probably won't be merged until I also implement https://github.com/atom/tabs/issues/275#issuecomment-191879765, to easier facilitate drag-and-dropping.~~
See GIF!
tab-bar-appearance-drag-and-drop

Along the way, I also fixed some very incorrect test assertions about the alwaysShowTabBar setting.

### Alternate Designs

None.

### Benefits

More consistent experience when "Always Show Tab Bar" is disabled.

### Possible Drawbacks

This removes the ability to drag tabs out of a pane that only has one tab.

### Applicable Issues

Fixes #275
Inspired by #277
Needs atom/tree-view#1293 to work correctly with the Tree View


lib/tab-bar-view.coffee
```diff
@@ -230,15 +237,12 @@ class TabBarView
   getWindowId: ->
     @windowId ?= atom.getCurrentWindow().id
 
-  shouldAllowDrag: ->
-    (@paneContainer.getPanes().length > 1) or (@pane.getItems().length > 1)
```
daviwil: Looks like there was logic for pane containers here too, how will they be affected by this change?

50Wliu: The logic for pane containers was what caused #275 to begin with. It made the tab bar always show if there was more than one pane container present.

daviwil: Cool, that sounds fine to me!

daviwil:
Hey @50Wliu, looks like a recent commit you made created a merge conflict that needs to be reconciled with the changes in this PR:

https://github.com/atom/tabs/commit/d6947e37a01573877a5b1be1647cabe6efa74e55

If you can update the PR to take care of that conflict, feel free to merge it afterward.  Thanks a bunch!

Issue #545, opened 2018-12-11
Middle click doesn't close tab


AlexWayfer:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

[Description of the issue]

### Steps to Reproduce

1. Click on tab via middle mouse button (wheel)

**Expected behavior:** Tab is closing.

**Actual behavior:** Tab isn't closing.

**Reproduces how often:** 100%

### Versions

Arch Linux.

```
> atom --version
apAtom    : 1.32.2
Electron: 3.0.10
Chrome  : 66.0.3359.181
Node    : 10.2.0

> apm --version
apm  2.1.3
npm  6.4.1
node 11.4.0 x64
atom 1.32.2
python 2.7.15
git 2.20.0
```

### Additional Information

Probably it appeared after upgrade to Electron 3.

Atom doesn't start with Electron 2 now.

Arcanemagus:
Thanks for reaching out!

We've determined that you are currently using an unofficial build or distribution of Atom. Often these customized versions of Atom are modified versions of the Stable branch of Atom with mismatched versions of built-in components. These updated components are taken from the Beta channel or `master` branch and then injected into the Stable version and a new Atom package is generated. Because of the way Atom is constructed, using these mismatched components can cause mysterious and hard-to-diagnose problems. You can find out more about [why we chose to not support unofficial distributions here](https://github.com/atom/design-decisions/blob/master/unofficial-atom-distributions.md).

You can find instructions for installing an official version of Atom in the [Flight Manual](https://flight-manual.atom.io/getting-started/sections/installing-atom/). If you are still seeing this problem on an official build please file a new issue, thanks!

PR #552, opened 2018-12-21
Fix tab close on middle click for Electron 3

VarNepvius:
### Description of the Problem

When using Atom with Electron 3, a middle click does not close a tab. This happens, because Electron 3 is based on a Chrome version >= 55. Since then, Chrome changed how click events work. Non-primary clicks now fire the event `auxclick` rather than `click`.

See https://developers.google.com/web/updates/2016/10/auxclick and https://developer.mozilla.org/en-US/docs/Web/Events/auxclick for more information on the event.

### Description of the Change

This PR adds the `onClick` handler of TabBarView also as listener to the `auxclick` event.

### Alternate Designs

None

### Benefits

The change fixes the problem for Electron 3 while it does have no effect in Electron 2, since the `auxclick` event never fires in this version. 

This helps with a future migration of Atom to Electron 3 and fixes the issue for unofficial distributions using Electron 3 now.

### Possible Drawbacks

None

### Applicable Issues

fixes atom/tabs#545

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -87,6 +87,7 @@ class TabBarView
 
     @element.addEventListener "mousedown", @onMouseDown.bind(this)
     @element.addEventListener "click", @onClick.bind(this)
+    @element.addEventListener "auxclick", @onClick.bind(this)
     @element.addEventListener "dblclick", @onDoubleClick.bind(this)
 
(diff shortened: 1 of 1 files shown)


daviwil:
Thanks a lot, @VarNepvius!  Verified that this fixes the issue on a build of Atom on Electron 3 and it also doesn't have any negative side effects on Atom 1.34.0 stable.

PR #554, opened 2019-01-12
Make item scroll tests more legible

50Wliu:
### Requirements

### Description of the Change

It really wasn't clear what these tests were doing before, which was partly because the width of each tab was not explicitly defined. The tests themselves were also wrong - `scrollIntoView` doesn't do anything if the item is at least partially visible. I've updated these tests to properly test that scrolling works as expected, and also updated `scrollIntoView` to use the new options listed at https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView.

### Alternate Designs

None.

### Benefits

The purpose of these tests should be more clear and the intended behavior more easy to follow.

### Possible Drawbacks

None.

### Applicable Issues

None.

### DIFF
--- a/lib/tab-bar-view.coffee
+++ b/lib/tab-bar-view.coffee
@@ -168,7 +168,7 @@ class TabBarView
       @activeTab?.element.classList.remove('active')
       @activeTab = tabView
       @activeTab.element.classList.add('active')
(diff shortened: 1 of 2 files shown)


50Wliu:
Ah yes, good old "it works on my laptop now but not on CI".

PR #560, opened 2020-11-04
Fix subfolder git detection

utkarshgupta137:
# Identify the Bug

Same as https://github.com/atom/tree-view/issues/1366

### Description of the Change

By using atom.project.repositoryForPath function instead of matching the path with added directories, subfolders will get identified.
Made necessary changes to referenced places to handle the async nature of the atom.project.repositoryForPath.

### Alternate Designs

None.

### Possible Drawbacks

Since atom.project.repositoryForPath is an async function, there maybe a little lag when loading file which already has changes.

### Verification Process

TODO

### Release Notes

- Fixed subfolders of project folders not getting detected as git repositories.

### DIFF
--- a/lib/mru-item-view.js
+++ b/lib/mru-item-view.js
@@ -1,10 +1,10 @@
 'use babel'
 
+import {Directory} from 'atom'
 import getIconServices from './get-icon-services'
-import path from 'path'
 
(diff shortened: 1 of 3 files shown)


utkarshgupta137:
TODO: Add tests.

